PR #17, opened 2015-09-09
Update installing-the-trial-version-of-mmc.adoc

smehgal:


seanosterberg:
Thanks for fixing this typo, we appreciate it.

PR #20, opened 2015-09-14
Trying to fix images broken link

mquintela:


seanosterberg:
Hi Mariano, you don't need to add the `_images` path in front of images. It looks like the image is appearing -- can we consider this closed?

PR #23, opened 2015-09-17
Fix assumptions in Clusters

eddumelendez:


seanosterberg:
Thank you for fixing these tables!

PR #39, opened 2015-09-24
Add release notes

IanWhyte:
Correcting previous request


seanosterberg:
Based off an old and removed commit.

PR #47, opened 2015-09-26
Release notes additions and fixes

IanWhyte:
These should work, but please double-check to verify no files are deleted


IanWhyte:
Add latest release note links to ToC file

PR #67, opened 2015-10-06
Change operation name in quick start

federicoboerr:


seanosterberg:
Thank you for the clarification, we appreciate it!

PR #75, opened 2015-10-12
Fix typos in v3.6 Auth HTTP Requests docs

baopham:


seanosterberg:
Thank you for these fixes, Bao! We appreciate it.

PR #77, opened 2015-10-12
Add anypoint-b2b-overview and anypoint-partner-manager-user's-guide t…

IanWhyte:
…o ToC


IanWhyte:
ToC changes to reflect added pages

IanWhyte:
Thanks - I apologize, my last iteration of link fixes was so incomplete
that I think I may submitted an interim save instead of what should have
been the final version (that is, I never saved the final version in Atom);
I've corrected and submitted
https://github.com/mulesoft/mulesoft-docs/pull/78.

Again, we're totally up for putting together a staging server here if that
makes sense to you.

Ian

PR #88, opened 2015-10-16
Urgent security fix + APM User's Guide content

IanWhyte:


IanWhyte:
Urgent security fix (+APM editing)

PR #98, opened 2015-10-20
Fix typo on Log4j2 configuration TIP.

f-lopes:
Fix Log4j2 configuration guide link.


seanosterberg:
Thanks for fixing the link and for making our docs better!

PR #99, opened 2015-10-21
Update creating-or-disbanding-a-cluster.adoc

jmjdamen:
Added information on the availability of the *New Cluster* option, which isn't available when the correct license isn't applied to one or more of the available servers.


george-hoffer:
Thank you for helping improve MuleSoft documentation - much appreciated.

PR #118, opened 2015-11-04
Update installing-the-production-version-of-mmc.adoc

jmjdamen:
for windows the contents of the setenv.bat file should start with set, not setenv, that won't work!


pedroielpi:
Thanks for spotting the error in setting JAVA_OPTS in Windows; contribution merged.

PR #130, opened 2015-11-05
Update installing-anypoint-enterprise-security.adoc

adrianmartinmulesoft:
Changed the Studio update site since it change from version to version and pointed to the release notes, where I defined them.


george-hoffer:
Looks good. Thanks for providing this.

PR #132, opened 2015-11-06
APMUG content additions

IanWhyte:


george-hoffer:
Thank you for handling this.

PR #134, opened 2015-11-06
Update ftp-transport-reference.adoc

jmjdamen:
Filename wildcard filter syntax fixed


george-hoffer:
Thank you for helping improve MuleSoft documentation.

PR #135, opened 2015-11-06
Updating to Anypoint Runtime Manager

aaronlandgraf:
The capabilities discussed here map directly into Anypoint Runtime Manager, which is a subset of Anypoint Management Center.  
- removed all references to Anypoint Management Center
- updated references to Anypoint Runtime Manager to reflect latest positioning


george-hoffer:
Thanks for the fixes.

PR #142, opened 2015-11-09
Update the link to the certification tab rather than the overview tab

NathanNam:


nearnshaw:
Thanks for contributing!

PR #144, opened 2015-11-10
Update request-reply-scope.adoc

adrianmartinmulesoft:
Clarify the default value for the timeout as it's not clear.


george-hoffer:
Looks good. Thanks for handling these docs.

PR #152, opened 2015-11-13
Remove spurious characters, fix links

IanWhyte:


george-hoffer:
Thank you for fixing these.

PR #155, opened 2015-11-17
Update configuring-mule-stacktraces.adoc

adrianmartinmulesoft:
Added the mule.verbose.exceptions since is the suggestion make in the logged exception:

i.e from log exception:
+ 3 more (set debug level logging or '-Dmule.verbose.exceptions=true' for everything)


george-hoffer:
Thank you for fixing this.

PR #157, opened 2015-11-18
Fixing the "copy objects" section

mcoaker:


george-hoffer:
Thank you for fixing this.

PR #159, opened 2015-11-18
Fix image links, incorp tech review

IanWhyte:


george-hoffer:
Thanks for getting these fixed.

PR #163, opened 2015-11-19
Spelling Error

jessefulcher05:
NetSuite connector is a operation based connector, which means that when you add the connector to your flow, you need to configure a specific web service for the connector to perform. The NetSuite connector v5.0 supports 56 operations.

*Net Suite connector is AN operation based connector

### DIFF
--- a/mule-user-guide/v/3.7/netsuite-connector.adoc
+++ b/mule-user-guide/v/3.7/netsuite-connector.adoc
@@ -201,7 +201,7 @@ For this mapping, start by generating a token using the SSO Kit provided by NetS
 
 == Using the Connector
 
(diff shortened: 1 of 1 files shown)


george-hoffer:
Thank you for your contribution. Please complete the contributor's agreement as described on https://docs.mulesoft.com/#making-a-contribution to register as a contributor so that we can accept your pull request.

jessefulcher05:
Thanks. I just registered and accepted the agreement.

nearnshaw:
Thanks Jesse for your contribution and for going through all of the trouble of accepting the agreement! We're working on making that workflow simpler in the future. 
We can now accept your pull request!

PR #165, opened 2015-11-20
Update script-component-reference.adoc

adrianmartinmulesoft:
- Added section `Script Properties included by default`
- Corrected a type with `Studio`


george-hoffer:
Thank you.

PR #175, opened 2015-11-23
Standardize release notes version-naming

IanWhyte:
Found a link-checking tool... And found this broken link, now fixed.


nearnshaw:
Thanks Ian!
I also adjusted the page name in the TOC file, otherwise our TOC won't be linking to this new URL after this change

IanWhyte:
Thanks - that makes sense, but the link-checking software wasn't finding
anything broken on our staging site, which is a little concerning - given
what you're saying, it should be. I'll look into it and follow-up if
appropriate.

Ian

PR #176, opened 2015-11-24
Update defining-slas-and-alerts.adoc

jmjdamen:
Fixed Broken Script Context Bindings URL


nearnshaw:
Thanks jmjdamen!
That link was pointing at the old docs site, and the URL is no longer valid. Thanks for pointing it to where our current docs are

PR #180, opened 2015-11-30
Applied the proper name (DevKit Studio Plugin)

NathanNam:


george-hoffer:
Thank you for creating these release notes

PR #181, opened 2015-11-30
Updated summary and updated libraries

asequeira:


george-hoffer:
Thank you for the corrections.

PR #188, opened 2015-12-02
ConnectionStrategy renamed to Config in Devkit 3.7.X or higher

juandesi:


george-hoffer:
Thanks for the doc fixes.

PR #198, opened 2015-12-09
Update building-an-external-oauth-2.0-provider-application.adoc

avaudagna:
mule-domain-config.xml is located in /domains/api-gateway. There is nothing in /domains/default.


george-hoffer:
Hi Alan - Thanks for submitting this fix.

PR #199, opened 2015-12-09
Update api-v2.raml

avaudagna:
In the AES Tutorial the /validate resource is configured on the 8083 port


george-hoffer:
Thank you for fixing this.

PR #211, opened 2015-12-12
Added the enabled flag.

wpoch:


nearnshaw:
I made these same changes separately in a different commit

PR #218, opened 2015-12-14
Update setting-up-external-identity.adoc

adrianmartinmulesoft:
Clarification about the Audience field


george-hoffer:
Thank you for handling this improvement.

PR #219, opened 2015-12-15
Update installing-mule-agent.adoc

adrianpineyro:
Removing 'contact support' link, as it is already documented.


george-hoffer:
Thank you for this correction.

PR #225, opened 2015-12-16
Pages and ToC entries for GA relates

IanWhyte:


george-hoffer:
Thanks for handling these updates.

PR #229, opened 2015-12-17
Explaining how to avoid Maven deployment

asequeira:


george-hoffer:
Thank you for providing this fix.

PR #231, opened 2015-12-18
Update http-basic-authentication.adoc

estebanwasinger:
* Changed from @ConnectionStrategy to @Config
* Added deprecated, due that this feature in DevKit 3.8.0 has gone deprecated.


george-hoffer:
Thank you Esteban!

PR #237, opened 2015-12-30
Added CSV Output Modifiers

PabloCamino:
I have added a table with optional CSV Output modifiers for DataWeave as they were not documented elsewhere.
This is handy information that users would like to have.
I was missing some defaults information, like the default character set or buffer size.
This pull request is associated to documentation JIRA issue DOCS-1478


nearnshaw:
Thanks Pablo, this information will surely be very valuable to many users :)

PR #239, opened 2016-01-04
Update munit-maven-support.adoc

diegostrubolini:


george-hoffer:
Thanks for fixing that.

PR #240, opened 2016-01-04
Replace “ quotes with machine-readable " character

conorcurlett:
Copy-pasting XML from the documentation will not be recognized by Anypoint Studio due to the use of the wrong quote character. This potentially is happening in other places in the documentation (a global search/replace would resolve).


george-hoffer:
Thank you. We'll see if we can find more of these.

PR #243, opened 2016-01-07
Example correction, and clarifying this will only work on Windows.

adrianpineyro:


george-hoffer:
Thank you for fixing this.

PR #244, opened 2016-01-08

adrianmartinmulesoft:
Corrected the escape character by default


nearnshaw:
Thanks for that correction Adrian!

PR #246, opened 2016-01-10
fixed invalid github.io URL

jmjdamen:


nearnshaw:
Thanks for noting that link was wrong Jmjdamen! The URL we originally had there was old, but we aren't using the other one (http://mulesoft.github.io/data-weave) either. We want to keep all of our users within the official docs page, so I'll now change the link to https://docs.mulesoft.com/mule-user-guide/v/3.7/dataweave-reference-documentation

PR #247, opened 2016-01-11
Update setting-up-external-identity.adoc

adrianpineyro:
Adding/Clarifying generic values required to configure SSO.


george-hoffer:
Thank you.

PR #249, opened 2016-01-11
URL needs to be fixed

jmjdamen:
URL pointed out should go to: https://docs.mulesoft.com/mule-user-guide/v/3.7/object-store-module-reference ,but adds .adoc at the end.
Don't know the fix, maybe two exist?


nearnshaw:
This is indeed very odd!  I never came across this bug before. As you say, the URL shouldn't end in `.adoc`, and everything seems to be well written out

nearnshaw:
There was a bug in our build process, the issue is now fixed and the link works well!
Thanks for bringing this to our attention! :)

PR #251, opened 2016-01-12
Fixed text formatting on xpath.adoc

karlssonsimon:
Added some spaces that had been lost.


seanosterberg:
Thank you for the fix.

PR #253, opened 2016-01-12
fixed formatting of examples

karlssonsimon:
Made last edit a bit to quickly, had to redo and be more thorough :). Also noted and fixed some other lost spaces.


nearnshaw:
Thanks!

PR #254, opened 2016-01-13
DataWeave reference change on input directive

PabloCamino:
Input directive optional parameters for CSV do not work any more.
They have been replaced by XML reader-property elements.
I have made the change.


nearnshaw:
Thanks!

PR #257, opened 2016-01-18
Update index.adoc for MUnit 1.1.0 

julianpascualMuleSoft:
Update on MUnit compatibilities to explain which version of Mule MUnit Support Version is compatible with ESB versions.


nearnshaw:
Thanks Julian!! That's very helpful!

PR #258, opened 2016-01-18
Added links to DecimalFormat and DateTimeFormatter

dfernandez79:
An explanation of accepted patterns in `{format: ... }` is missing on the docs. That feature it's based in Java classes; the link to JavaDocs was added.


nearnshaw:
Thanks Diego!  Very helpful contribution!

Just an FYI, when adding links in asciidoc, you need to add the word "link:" before the actual link
(for example link:https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html[DateTimeFormatter])
I'll fix that , don't worry

dfernandez79:
Thanks didn't know about the link: prefix

PR #263, opened 2016-01-19
Updated docs for APIKit OData Extension v0.9.1

cazala:
Updated docs for APIKit OData Extension v0.9.1 (APIKit v1.7.3)


george-hoffer:
Thank you Juan. The fix will appear on the live site in about an hour.

PR #268, opened 2016-01-21
Update walkthrough-design-new.adoc

jmjdamen:
correction in version number, did not match with earlier walkthrough steps


nearnshaw:
Thanks Joost!

PR #269, opened 2016-01-21
Syntax fixes for Dataweave

jmjdamen:
Quite some changes on wrong mappings, auto scaffolding and input data.
Dataweave scripts changed, walkthrough instructions needs to be updated


nearnshaw:
Thanks Joost!  There are some very helpful corrections here!  :)
Just an FYI: in asciidoc, exclamation marks don't comment text out. I'll address your comments now, so no worries, but for future reference, please comment text out with //  or ////
http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#comments

PR #270, opened 2016-01-21
[walkthrough-build.adoc] Update step 15.

fermujica:
Give a clearer explanation on where to locate the transform section, since the UI does not refer to this section as so.


nearnshaw:
:)

Issue #262, opened 2016-01-19
DevKit Setting up Environment should demonstrate how to set JDK as installed JRE within Eclipse


JoshRosso:
The JDK section within DevKit Setting Up Your Development Environment (https://docs.mulesoft.com/anypoint-connector-devkit/v/3.7/setting-up-your-dev-environment#jdk-7) should demonstrate how to link JDK to Studio.

Very often Studio users have the JRE within Studio's Installed JREs when they actually need their JDK home.

When building DevKit projects, they then receive the following error.
```
Unresolveable build extension: Plugin org.mule.tools.maven:mule-app-maven-plugin:1.1 or one of its dependencies could not be resolved: Could not find artifact com.sun:tools:jar:1.5.0 at specified path.
```

JoshRosso:
Issue closed from merge #275.

PR #275, opened 2016-01-23
Add details on setting up Installed JREs in Eclipse to support DevKit env config

JoshRosso:
Per issue #262.


george-hoffer:
Excellent changes. Thank you Josh.

PR #279, opened 2016-01-27
Update mule-expression-language-date-and-time-functions.adoc

germansolis:
Changing the method definition:
plusDay(int add)

with the correct one:
plusDays(int add)

The final 's' was missing, in this page there are examples of the use of that function with the right sintaxis


nearnshaw:
Thanks German!

PR #286, opened 2016-02-02
Fixing default reference

afelisatti:
Moved the default tip to the right place and added protocol to the Behaviour section. Feel free to edit.


nearnshaw:
Thanks Anita!!

PR #300, opened 2016-02-11
update from .bundle to .zip file

damimj:
.bundle is no longer the file extension for exporting APIs.


george-hoffer:
Thank you.

PR #309, opened 2016-02-12
Analytics chart removed from API Platform dashboard (landing page)

cesaraugustogarcia:
As part of the removal of Analytics chart, also the way to go to Analytics site has changed. Now the user has to select *Analytics* on the top right menu. So the image under `Accessing the Analytics Dashboards` has to be updated.
The following image can be used to replace the mentioned one: ` image:AnalyticsLink.png[AnalyticsLink]`, if team feels that meets the docs standards.
analyticslink


george-hoffer:
Thank you.

PR #312, opened 2016-02-12
1.3.3 version is compatible since 3.5.4 on

asequeira:


george-hoffer:
Thanks for fixing this.

PR #317, opened 2016-02-16
Update microsoft-dynamics-crm-connector.adoc

federicoboerr:
Connector now runs on Java 8 and doc got outdated.


nearnshaw:
Thanks Fede!

PR #318, opened 2016-02-16
fix typo in xml example

damiansima:


george-hoffer:
Thank you for fixing this.

PR #321, opened 2016-02-18
Update installing-anypoint-runtime-manager-on-premises-edition.adoc

germansolis:
The current enablement process for the on-prem platform is not done via support, but pre-sales or services, I've updated the documentation to reflect this


fermujica:
Thanks for your contribution, German!
I'm merging your commit to properly reflect this in our docs.

PR #323, opened 2016-02-21
grammer change: for -> from

ethanport:


nearnshaw:
Thanks Ethan, agreed!

PR #326, opened 2016-02-21
Add warning: don't register both MMC and hybrid

ethanport:


nearnshaw:
Thanks Ethan! It's a very valid point!

PR #327, opened 2016-02-21
Fix of cron examples

mauro-flores:
Original examples were not using a proper cron notation, "human description" didn't map properly the cron expression. Changes have been validated with cron translators and cron expression validators.


george-hoffer:
Thanks for handling this.

PR #330, opened 2016-02-25
Add multi value selector to filter

brechtvdv:


george-hoffer:
Thank you for catching this.

PR #331, opened 2016-02-25
Clarify use case when API Gateway is deployed in CloudHub

eguser:


george-hoffer:
Thank you for catching this.

PR #335, opened 2016-02-29
Added example of using MuleEventContext

adobni:
instead of using MuleClient for section When Not to Use the Mule Client


george-hoffer:
Thank you Alejandro.

PR #336, opened 2016-02-29
Incorrect spelling 'flowss'

adobni:


george-hoffer:
Thanks Alejandro. There was a branch conflict so I made the fix directly. Should appear within an hour.

PR #338, opened 2016-03-01
Fixed RAML 0.8 spec references

kjervis:


fermujica:
Great catch! Thanks for fixing it!

PR #339, opened 2016-03-02
Reorganize content, add prod warning

ethanport:
Reorganized content. Added warning against Hot deployment in production.


george-hoffer:
Thanks Ethan!

PR #342, opened 2016-03-04
add links to mulesoft github projects

ethanport:


nearnshaw:
Thanks Ethan!
I'm replicating these same changes on the 3.8-m1 version too

PR #343, opened 2016-03-04
add links to source code for additional modules

ethanport:


nearnshaw:
Sorry Ethan... I had seen you did these changes on 3.7 and I copied them on 3.8-m1 right away without seeing that you had already done those on a separate pull request!
I will close this one so that we don't have conflicting commits, but rest assured that the source code now includes these same changes you made here.
I really appreciate that you go the extra mile with these things :)

PR #344, opened 2016-03-05
Use full xml example property-placeholder

JoshRosso:
Use full xml example property-placeholder. The current xmlns placement isn't clear to a non-spring user.


george-hoffer:
Thanks, Josh!

PR #354, opened 2016-03-08
edit the version since which coverage features is available from 1.2.…

damiansima:
…0 to 1.1.0


george-hoffer:
Thank you for fixing this!

PR #355, opened 2016-03-08
add ignore flow configuration example

damiansima:


george-hoffer:
Thanks for adding this.

PR #357, opened 2016-03-08
Added Authorization Filter

PabloCamino:
Authorization filter, part of the _mule-ss_ schema was totally absent of the documentation.
Also, updated the security filters code lines to use the new _http:basic-security-filter_ and be consistent with the main example above.


george-hoffer:
Thank you.

PabloCamino:
Hi @george-hoffer this update got lost with the last documentation release.
Would you like to make the changes for 3.8 documentation?
Best.

PR #361, opened 2016-03-10
Andes has never been a supported release.

asequeira:


nearnshaw:
Thanks Ale! Yes.. at one point when it was a new beta it made sense to have that release notes doc there, but now no one is going to use that old beta version, keeping it just confuses people.

PR #366, opened 2016-03-12
Removing "Modifying the Cleanup Script for MySQL"

adrianpineyro:
This bug was fixed as part of the the MMC 3.5.0 release; there is no need to use this script starting from MMC 3.5.0. More info at https://www.mulesoft.org/jira/browse/SE-373


fermujica:
Thanks for the heads up! 
So, this whole paragraph should also be removed from  `/v/3.5/persisting-mmc-data-to-mysql.adoc` and `/v/3.6/persisting-mmc-data-to-mysql.adoc` as well, right?

PR #371, opened 2016-03-15
Removed references to Anypoint Runtime Manager

aaronlandgraf:
As Anypoint Runtime Manager is not GA (and will not be until now November), this needed to be updated.  Also fixed a few minor branding issues that I caught during my review.


nearnshaw:
Thanks Aaron!
There was initially a great deal of confusion about if the Runtime Manager on premises was at a stage where we could openly advertise it or not. Thanks for making these changes!

PR #373, opened 2016-03-15
Add new compatibility rules for MUnit 1.2.0

diegostrubolini:


george-hoffer:
Looks good. Thanks for fixing these.

PR #379, opened 2016-03-18
Fix code tags

estebanwasinger:


estebanwasinger:
Can be cherry picked to 3.7

PR #381, opened 2016-03-21
added Maven troubleshooting and links

ethanport:
added Maven troubleshooting and links


fermujica:
Thanks Ethan, this is great. There is just one quick fix:
> Studio includes a Maven Support view. You can add this to your workspace projects by selecting Window > Show View > Other..., then searching for Maven Support. 

This is only possible if the user has _*Maven Tools for Mule*_ (not included in the default installation) installed. I'm linking the instructions for that as well.

PR #390, opened 2016-03-23
Added trim operator.

adobni:
With definition and example.


nearnshaw:
Thanks Ale!  I didn't know that operator existed, great that you added it!

PR #395, opened 2016-03-25
A stray single 's' was removed (this caused invalid json).

cgorshing:
The very first example had a stray 's' for the input JSON data, I copied/pasted a few times thinking I was doing something wrong until I saw the 's'. I expected the JSON to be valid - hopefully I'm rightfully so in thinking :)


george-hoffer:
Thank you for catching this, and thanks for accepting the contributor's agreement. Appreciate your help in improving MuleSoft documentation.

PR #396, opened 2016-03-25
Add yourkit agent instructions via -attach method.

JoshRosso:
I believe we should add instructions on how to link the YourKit agent into Mule via `-attach`. It's quite convenient.


george-hoffer:
Wow - Very nice documenting. Thank you!!

PR #397, opened 2016-03-26
Remove _images ref from images and fix anchor links to be consistent

JoshRosso:
Remove _images ref from images and fix anchor links to be consistent with how the site build.

@george-hoffer A few mistakes I made in refing links and images. Corrected here.


nearnshaw:
Thanks Josh!
Both things definitively needed fixing. I'm not sure though that the link:#internal-link-to-title  syntax really works in asciidoc, at least I've never used it. Does it? For those cases, what I use is the following syntax <<this is the title to link to, this is the text to display in the hyperlink>>
I'll approve your pull request and then change that

PR #398, opened 2016-03-28
CloudHub VPC

mhuang55:
Cleaning up the wording on VPC and to separate VPC from the connectivity methods such as IPSec.

### DIFF
--- a/cloudhub/v/latest/virtual-private-cloud.adoc
+++ b/cloudhub/v/latest/virtual-private-cloud.adoc
@@ -9,7 +9,7 @@ The Virtual Private Cloud creates a logically private and isolated network to ho
 
 image:CHVPC01.png[CHVPC01]
 
-VPC can be connected to your corporate intranet or cloud infrastrcture using secure connectivity methods. You can leverage IPSec (Strongswan) or Amazon Web Services (AWS) Direct Connect to connect to remote networks. In the event that your remote network is also in AWS, VPC peering can be leverage given your CloudHub VPC and your AWS VPC are in the same region. Do note that to leverage AWS Direct Connect you must have the relationshiop with AWS to set up the Direct Connections. Once the direct connection is established it can be applied to your CloudHub VPC. Although there are serveral choices highlighted here, IPSec is in gernal the recommended solutions as it provides a secure and flexible way to connect your VPC to your on-prem networks.
+VPC can be connected to your corporate intranet or cloud infrastrcture using secure connectivity methods. You can leverage IPSec (Strongswan) or Amazon Web Services (AWS) Direct Connect to connect to remote networks. In the event that your remote network is also in AWS, VPC peering can be leverage given your CloudHub VPC and your AWS VPC are in the same region. Do note that to leverage AWS Direct Connect you must have the relationshiop with AWS to set up the direct connections. Once the direct connection is established it can be applied to your CloudHub VPC. Although there are serveral choices highlighted here, IPSec is in gernal the recommended solutions as it provides a secure and flexible way to connect your VPC to your on-prem networks.
 
 == VPC Connectivity Methods
 
--- a/cloudhub/v/latest/virtual-private-cloud.adoc
+++ b/cloudhub/v/latest/virtual-private-cloud.adoc
@@ -1,7 +1,7 @@
 = Virtual Private Cloud
 :keywords: cloudhub, vpc, ipsec, ssl, openvpn
 
-CloudHub's Virtual Private Cloud (VPC) allows you ro create a virtually private and logically isolated network in the could to host your integration work loads.  You have the choice of using several connectivity methods that can connect your VPC securely and privately to your corporate intranet – whether on-premises or in other clouds – as if they were all part of a single, private network. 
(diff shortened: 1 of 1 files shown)


fermujica:
Thanks a lot for this Mark!  
Last week we were revising this very VPC definition as it was confusing to some users.
As we are trying to clarify our VPC offering a little bit more, I'll definitely use your suggestions in this PR in our [current draft](https://github.com/mulesoft/mulesoft-docs/blob/Runtime-Manager-rename-2016/runtime-manager/v/latest/virtual-private-cloud.adoc) .

mhuang55:
So a few things ... one let's remove OpenVPN as this is not something we're going to support going forward. Most of our configurations are IPSec and there's no need for OpenVPN (SSL) since IPSec does the trick. Also those are not topologies. A VPC is AWS's term for their network virtualization layer and that's it. It simply exposes an API so that you can create a network. To connect this network to any other network has nothing to do with VPC. It has everything to do with HOW one wants to connect. So those are connectivity methods or models. It is also critical important to highlight that a VPC in the way we use it is an extension of the customer's network. They define the address space that they can route over their intranet. This is a critical point as it causes much of our grief in the initial conversations with customers. We have to un tech the notion that we define VPC and how to connect. The fact of the matter is, this is part of the network design.

fermujica:
> So a few things ... one let's remove OpenVPN as this is not something we're going to support going forward. Most of our configurations are IPSec and there's no need for OpenVPN (SSL) since IPSec does the trick. 

Got it. Not sure if we should remove it completely or let them know that this is a possibility that we don't recommend and are unnoficially sunsetting. 

> Also those are not topologies. A VPC is AWS's term for their network virtualization layer and that's it. It simply exposes an API so that you can create a network. To connect this network to any other network has nothing to do with VPC. It has everything to do with HOW one wants to connect. So those are connectivity methods or models. 

Totally agree. I think this _topologies_ denomination along with the whole deal of portraying VPC as a product to connect networks was more of a marketing approach, similar to what Daniel said in his [comment](https://mulesoft.slack.com/archives/docs-feedback/p1458525930000039). 

>It is also critical important to highlight that a VPC in the way we use it is an extension of the customer's network. They define the address space that they can route over their intranet. This is a critical point as it causes much of our grief in the initial conversations with customers. We have to un tech the notion that we define VPC and how to connect. The fact of the matter is, this is part of the network design.

Yes, I understand the need to properly define a VPC.  I'm rewriting this part to make it clearer. Expect an update on [the draft](https://github.com/mulesoft/mulesoft-docs/blob/Runtime-Manager-rename-2016/runtime-manager/v/latest/virtual-private-cloud.adoc) later today. We are currently handling most of the updates for Cloudhub related docs in a separate branch so we can properly migrate all `Cloudhub` directories to `runtime-manager` definitions in one simple merge. 
If these changes do not affect the current branch, I'd like to push these updates to `Master` so they can go live already. I'll keep you posted.

I've been revising our VPC doc for a week now. There's even JIRA [DOCS-1543](https://www.mulesoft.org/jira/browse/DOCS-1543) created for this, so I really appreciate your comments. They **really** helped.

fermujica:
Hi Mark,

As we talked earlier, I organized our VPC documentation.
* The *topology* definition has been removed and replaced by *Connectivity Method* as you suggested in this pull request. 
* I checked the implications of removing the openVPN option from our documentation and luckily there are no other references to this technology except for the VPC discovery form where it still lists 
**OpenVPN tunnel with network-to-network configuration** and **OpenVPN tunnel with host-to-network configuration** as valid connection modes. I don't have access to modify it, but if openVPN is being taken out of the documentation, this form should be updated as well. 

Check the [current draft](https://github.com/mulesoft/mulesoft-docs/blob/Runtime-Manager-rename-2016/runtime-manager/v/latest/virtual-private-cloud.adoc) to see the final version and let me know what you think.

mhuang55:
So I looked at the current draft: And here are things I think should be cleaned up still.
In how VPC works .. it still talks about VPC topologies. VPC is just a service to create a network. The wording also says VPC can be configured to use IPSec .. we can't actually configure a VPC in that way. We configure an IPSec GW to leverage IPSec. I'm not sure what "IPSec can be configured at the HW level in addition to SW client" is intended to tell the reader. And the wording on peering seems to be prescribing how something should be done. The fact of the matter is that peering is just an option even if the customer is in AWS and in the same region. This should not be an instant recommendation that peering should be used. So even if OpenVPN is still in the form which we'll need to get that all sorted out in a more long terms strategy, educating the customer start with this doc. Most customer reference this doc when we question them on their approach. So far we've only gotten 2 OpenVPN requests that I'm aware of. When we ask why they need to use it, they just say the documentation said I could. We then tell them to use IPSec and they go okay. There's normally no compelling reason to use SSL (OpenVPN) unless there's specific use-case which we have not come across. Also we've gotten more request to link their direct connect than we have in OpenVPN. And in some instances the customer had no idea we even did this and after having a conversation with them, they would have wished to do that instead of IPSec but it was too late for them to change. Also we need to convey the point that the VPC and how to connect is defined by the customer not Mulesoft. This is also an area of contention and confusing as they think we have some default configuration. Since this is like an extension of their network, they actually defined it. We just do the typing for them. When all this becomes self services at some point we're not longer involved. They would have an interface to define and configure all this on their own.

PR #399, opened 2016-03-29
Correction of a few typo's

cgorshing:
Many typo corrections. Some I didn't change because although they are not proper USA spelling, they are correct British English spelling.

There were also some typo's in the release notes which were commit messages. I left those as is as well.


nearnshaw:
Thanks a lot Chad! You even made the effort of replicating changes across versions, we really appreciate that :)

PR #401, opened 2016-03-29
Add an additional argument for license generation command

NathanNam:


sharkberto:
Thanks Nathan. It is no problem to add this information there, in my view--I only have a minor adjustment after merge...

PR #403, opened 2016-03-30
payload.bookstore.book should be payload.bookstore

ethanport:
This example does not work without this change. The change generates the listed Output.

### DIFF
--- a/mule-user-guide/v/3.7/dataweave-examples.adoc
+++ b/mule-user-guide/v/3.7/dataweave-examples.adoc
@@ -271,16 +271,16 @@ This example takes an XML list of books and outputs a JSON version of the same. 
 %input payload application/xml
 %output application/json
 ---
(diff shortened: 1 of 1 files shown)


nearnshaw:
Thanks Ethan, you're right about that, the example is broken.
A syntax of  payload.bookstore.*book   (with an *) would also work, which is what we have in several other examples. Your version is easier to read and understand, though... so I'm actually going to change all of the others to match yours :)

nearnshaw:
Thanks Ethan, you're right about that, the example is broken.
A syntax of  payload.bookstore.*book   (with an *) would also work, which is what we have in several other examples. Your version is easier to read and understand, though... so I'm actually going to change all of the others to match yours :)

ethanport:
Actually, it would be great to include your alternative also. This syntax is confusing when you first learn it.

PR #405, opened 2016-03-30
Fix business group parameter

aiannucci:
Add the businessGroup parameter which was added in version 2.1.1 and was missing from the docs.


nearnshaw:
Thanks Ale!   I'll now also add this parameter to the 3.8-m1 version of that doc

PR #406, opened 2016-03-31
Update windows-gateway-services-guide.adoc

mmansur:
I found out from Federico Boerr that this pre-requisite is needed. However it is not documented here.


fermujica:
Great! Thanks for pointing this out. 
I'll add this comment to `/v/3.7` and `/v/3.8-m1` as well.

maurok:
Please review this. JCE is required at the machines where the connector is running, not the server where the gateway is installed.

PR #410, opened 2016-04-04
Added a TIP to server tasks

PabloCamino:
Adding a small tip box highlighting that if a server is running when registered it needs to be restarted for it to communicate with the runtime manager. It's not explicitly documented and a customer recently got stuck on this step because of this.


nearnshaw:
Thanks Pablo!
this page will soon be replaced by something I'm working on in a different branch, but I'll still copy your comment to that other doc to make sure it lives on. I'm merging your changes to the existing doc anyway.

PR #416, opened 2016-04-07
Add link to Mule Expression Language Tips page

adobni:


nearnshaw:
Thanks! I'll merge this change, and also add the same link in other versions of the doc :)

PR #426, opened 2016-04-11
Update anypoint-studio-may-2014-with-3.5.0-runtime-release-notes.adoc

adrianpineyro:
Changing the issue number, so it can be found by the public JIRA (EE tickets are private)


fermujica:
Thanks for changing this!

PR #434, opened 2016-04-14
Update in DataWeave for Studio input typre

PabloCamino:
On Studio, actually on Mule, the input directive does not work to define the input type.
Neither for payload nor for any kind of input.
In order to define the input type the mimeType attribute should be used on the appropriate input- elements.

See DOCS-1476


nearnshaw:
Thanks Pablo, merging it now to the doc

PR #449, opened 2016-04-19
Remove contraint not valid anymore

cesaraugustogarcia:
When the feature was firstly released it went out without images and attachment being included as part of export feature. On a next release this was added, so this is not a valid constraint anymore


fermujica:
I just merged it!
Thanks for the update.

PR #452, opened 2016-04-21
more relnote link-fixing

IanWhyte:


george-hoffer:
Thanks for fixing this.

PR #464, opened 2016-04-29
Update apikit-beyond-the-basics.adoc

tonedef71:
Added an example to show how one would handle create a mapping to manually handle the 500 Internal Server Error response in a custom way.  Such an example is relevant for generic handling of certain runtime exceptions that arise in DataWeave script that would proceed unreported otherwise.


krishahn:
Thanks tonedef!

PR #469, opened 2016-05-04
Update servlet-transport-reference.adoc

manibh:
We need to change following:
http://host:port/webapp-contextname/servlet-name/path
to:
http://host:port/webapp-contextname/servlet-mapping-url/path


george-hoffer:
Thank you, Mani!

PR #470, opened 2016-05-04
Update validations-module.adoc

brs-mule:
Added required Maven dependency for Maven projects.


fermujica:
Thanks for adding this.

PR #473, opened 2016-05-05
Corrected top level positioning on API entitlements

aaronlandgraf:
Removed confusing language RE: continuing user as per Matt Kilguss' request.  Also brought value messaging to the top.


george-hoffer:
Thank you for providing this fix.

PR #476, opened 2016-05-06
Improved description of the standalone value

dfernandez79:
The current description of the standalone parameter makes you think that `standalone: false` means that the policy has dependencies. That is not what happens. Standalone false means that the policy can only be used as part of another policy (similar to "abstract" in a programming language).

I tried to improve the description. Not sure if my use of english was good.


krishahn:
Thanks, Diego.

PR #477, opened 2016-05-06
Update applying-custom-policies.adoc

dfernandez79:
Removed the description of type in the policy YAML. Reason: type is optional, and supported options are system or custom. However system is used internally; and users are not able to upload custom policies with type: system. This document describes the creation of custom policies, so it will confusing for users.
I also removed the type: system from the example... doing a copy & paste of that example will not work.


krishahn:
Thanks Diego. I'll make sure your changes make it into the google doc for the upcoming release: https://docs.google.com/document/d/1PcKZBku73n0tI7bLgeRQObBRB8x6AuzkfZBVKZuhRJ8

Also, checking that the your changes made it into the docs after I resolved the merge conflict.

PR #479, opened 2016-05-10
Update Tomcat deployment instructions

adobni:
Current instructions are out of date for Mule 3.7.x and don't work. Probably don't work with 3.6.x either but I didn't try that. I tested this deploying a Mule web app in Tomcat 7 and Mule 3.7.3.


george-hoffer:
Thank you for fixing this!

PR #480, opened 2016-05-10
Using "indicate route" for "browse to a file"

cgorshing:
The wording of "indicate route" tripped me up as my head was deep into thinking about Dataweave and I thought this was some kind of DataWeave nomenclature.

I changed the wording to something I felt was more directed at selecting a file. Hopefully this makes it a bit easier to read and understand.

No worries if you all think otherwise.


nearnshaw:
Thanks Chad, agreed!

PR #481, opened 2016-05-11
Correction not carried over to 3.8-m1 side

cgorshing:
There is a correction in the 3.7 docs that changes:

    otherwise size: "XXL"
to

    otherwise "XXL"
But was not carried over to the 3.8-m1 branch.

### DIFF
--- a/mule-user-guide/v/3.7/dataweave-tutorial.adoc
+++ b/mule-user-guide/v/3.7/dataweave-tutorial.adoc
@@ -62,7 +62,7 @@ You then need to select only a few of these fields, rename some, and assume a va
 +
 image:add_metadatadw.png[image]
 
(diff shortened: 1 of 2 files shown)


nearnshaw:
Yes, you are absolutely right!
This tutorial you're editing is undergoing a lot of changes for the 3.8 release and will be renamed "dataweave quickstart", actually most of it will change. We'll be releasing this version a week from now. I'm taking all of your contributions into account for this new unpublished version I'm working on, just an FYI.

cgorshing:
Is https://docs.mulesoft.com/mule-user-guide/v/3.7/dataweave-reference-documentation also going through an over-haul as well? I'm working through this documentation now and I'm finding some things I'd like changed. I wasn't going to spend time on it though if it were to change in one weeks time.

PR #484, opened 2016-05-11

tonedef71:
Added descriptions and examples for two previously undocumented Dataweave functions: zip and unzip.


nearnshaw:
Thanks Tony!
For 3.8 we were actually going to include that operator in our doc that we're releasing on monday. But we still need to add this to older versions so very helpful!
I actually like your description a lot more, so if you don't mind I'll be adding a little of your words to the 3.8 version too :)

PR #487, opened 2016-05-11
Ms ps 1.6.0

maurok:


george-hoffer:
Thanks for handling these changes.

PR #489, opened 2016-05-13
Update anypoint-exchange2.adoc

brenda:


brenda:
I don't know if we want to clarify this, or if we want to have this document just thinking in the new version.

PR #491, opened 2016-05-13
Update anypoint-exchange2.adoc

brenda:


george-hoffer:
Thank you for making these changes.

PR #502, opened 2016-05-17
Update information about using MUnit on Studio

julianpascualMuleSoft:
Due that MUnit comes by default with Studio since 5.4.0, there is no need to explain how to install it at first. Just in case that they are using a previous version of Studio, we are leaving it at the bottom of the page just to inform them how to do it. This will need to be replicated on version 1.1.1 and 1.0.0


fermujica:
Good call. I did this for the quickstarts section, but by the time I updated MUnit articles forgot to pay attention to this matter.
Thanks for catching it.

PR #505, opened 2016-05-19
Auto-scaling isn't released.

dandiep:


nearnshaw:
Thanks Dan, I wasn't aware that was the case... I'll remove that topic from the TOC then too

PR #506, opened 2016-05-20
Incorrect version for 3.5.0 link

adobni:


nearnshaw:
Thanks for fixing it, that was a hard one to notice!!

PR #509, opened 2016-05-21
removed unnecessary DW-distraction from example

PaulAndersonMule:
Good documentation and examples focus on first demonstrating/explaining the core concept, and nothing else. In this case the objective is demonstration of use of object-store.

I added notes about exceptions that will occur, by design, when trying to re-store an existing value, or retrieve a nonexistent value.

The example has been stripped down to the bare essentials to ensure that users have as seamless, simple and direct an experience as possible.

Including DW here is not helpful to make the object-store behavior/usage clearer.


sharkberto:
Implementing these changes manually. Thank you for your contribution, Paul.

PR #510, opened 2016-05-23
removing unsupported Google connectors

adobni:
In Anypoint Exchange they are shown to be community supported so we don't guarantee they are maintained nor working with current releases.


sharkberto:
Thanks!

PR #511, opened 2016-05-23
Update first-30-minutes-with-mule.adoc

jbugnet:
There is a broken link in this page


nearnshaw:
Thanks!
I will fix the broken link you found, I can't accept your pull request as if I do, any changes you made to the content are directly built into the live site users see, and what you added was more of a note for us. But we will be fixing the issue nevertheless!

jbugnet:
No problem Nicolas, I was not sure how to just submit a comment. 

Jerome

PR #516, opened 2016-05-26
Fix version of dbserver dependency

julianpascualMuleSoft:
The dbserver dependency version is not the same as munit. It has a different release lifecycle.


fermujica:
Thanks, I wasn't aware of this.

PR #519, opened 2016-05-26
small workers are not only for test workloads. Removing phrases which are confusing customers. 

dandiep:


fermujica:
Thanks! Already merged this to our master.

PR #523, opened 2016-05-31
Update mule-high-availability-ha-clusters.adoc

ethanport:
Change Mule ESB, Mule instance, Mule server to Mule runtime, which is now the correct name.


fermujica:
Thanks for taking the time of going through this!

PR #525, opened 2016-05-31
Changing the version mentioned to 3.7

adobni:
because this is the doc for 3.7. Probably doc for 3.8 has the same issue.


nearnshaw:
Thanks Ale, I'll check version 3.8 to see if it has the same text

PR #537, opened 2016-06-03
on-premises, not on-premis

ethanport:


nearnshaw:
Hi Ethan,
Although there is controversy in this issue, the rule of thumb we're following with the whole on-premise / on-premises debate is the following:
* on-premises is to be used in direct cases, ie: my mule is running on-premises
* on-premise is to be used as an adjective that modifies something else, ie: I have an on-premise deployment
In the case of the "on-premise edition", we're considering that on-premise is an adjective that modifies edition
I'm therefore not merging your pull request, but let me know in case marketing is actually calling the prodcut differently.

PR #547, opened 2016-06-09
how queues and object stores are implemented

ethanport:
Added details on how vm persistent queues and object stores are implemented in CloudHub Fabric. 
Also added some details on HTTP load balancing and how to bypass the load balancer.


george-hoffer:
Thanks Ethan

nearnshaw:
Nice!  Thanks Ethan, what you've added reads out very clearly!

PR #548, opened 2016-06-09
Clarified minThreads, maxThreads

ethanport:
Clarified how these parameters work together. 

Should add a note on what happens when maxThreads is reached and maxBufferSize is reached. Does the connector deny subsequent requests?


nearnshaw:
Thanks Ethan, that paragraph really needed to be clarified, nice work!

PR #549, opened 2016-06-09
Incorrect support reference

adobni:
Corrected most links to support email. Customers should use the support portal. Inquiries about sales are directed to the sales inquiries email account.


george-hoffer:
Thank you Alejandro!

PR #552, opened 2016-06-10

tonedef71:
Minor typographical error.


nearnshaw:
Oddly there's a conflict with this pull request... sorry but I think the best way around this is to close your request without merging and then doing the same change separately.
Rest assured that your correction will be done on the doc, thanks for reporting it!

tonedef71:
That is odd.  Perhaps the system gets confused if merges are performed in a different order from that of the order that was committed?

PR #553, opened 2016-06-10

tonedef71:
Corrected a source type to be "DATAWEAVE" instead of "RUBY".  Also, lots of (unnecessary) updates to make DataWeave source declarations look more consistent (I can be anal-retentive sometimes).


nearnshaw:
haha ok, thanks Tony, also helpful!
Those code blocks all used to be marked as Ruby way back at some point, we missed that one apparently

PR #554, opened 2016-06-10
Fix JDK compatibiltiies

aiannucci:


nearnshaw:
Thanks Ale, merging now!

PR #555, opened 2016-06-10
Fix JDK compatibiltiies

aiannucci:


fermujica:
Thanks for fixing this!

PR #557, opened 2016-06-11
Fixed broken params table

ethanport:
Fixed broken params and clarified usage


nearnshaw:
Thanks Ethan!  Approving now!

PR #558, opened 2016-06-12
Clarified operations behavior

ethanport:
Added details on various operations behavior based on the source code. This change applies to older versions of the doc as well.


nearnshaw:
Nice! This is all very valuable stuff. I hear object stores tend to be a big source of confusion so we really appreciate that you're bringing in more clarity to that section :)

PR #559, opened 2016-06-13
Clarify objectstore operations ref

ethanport:
Added details to objectstore operations reference.


nearnshaw:
Again, thanks for contributing this!

PR #560, opened 2016-06-13
Document the custom log appender feature for CloudHub.

dandiep:


nearnshaw:
Great, thanks Dan!
I assume that this doc is for a still beta feature and not to be distributed openly, and that's why it's not listed in the TOC, right?

dandiep:
Correct.

PR #564, opened 2016-06-15
Removed tabs from code listing

cgorshing:
This code listing had some tabs intermixed with spaces and when viewed through the browser showed less than ideal.


nearnshaw:
Thanks Chad!
I'll do the same changes in the 3.8 version too. It does make a difference in readability, yup

PR #565, opened 2016-06-15
remove or supply correct update url's for web

jmjdamen:


nearnshaw:
Yes, all three of those URLs are outdated. I'll remove them, thanks!

PR #566, opened 2016-06-15
remove domain not supported waning from 1.2.0 version

damiansima:


fermujica:
I removed a couple of `Munit does not support mule domains` messages across the docs, but I seem to have missed this one.
Thanks for fixing it!

PR #569, opened 2016-06-16
Clarify fractional vCore workers intended usage

adobni:


fermujica:
Fair change. Thanks for clarifying that.

PR #570, opened 2016-06-17
DOCS-708

lee3star:
DOCS-708 customer's requesting for a detailed explanation on how to use salesforce headers


george-hoffer:
Thank you Sean

PR #574, opened 2016-06-20
DataMapper --> DataWeave

jmjdamen:
In Batch example DataWeave is used instead of DataMapper, updated text / various URL's for reference.


fermujica:
Indeed. Thanks for fixing it!

PR #583, opened 2016-06-24
result of readthrough, some corrections and ideas for better sentence…

sharkberto:
… and information flow, avoided using 'it', distilled information to keep sentences concise for readers


nearnshaw:
I really like these improvements, thanks!

PR #585, opened 2016-06-28
Add Gradle build script for automating full and single-page site build

mojavelinux:
A first attempt at a self-contained, fully-automated build. Note that this build currently only applies for a local development preview (intended for writers). The next step is to configure it to handle the CI build as well.

Instructions for how to use the build are included in the README.

The build currently draws the docs builder jar from s3. We definitely want to get this fixed soon so that it downloads a stable version from a proper repository.


mojavelinux:
:+1: 

I'll now proceed with the PR for adding the publish/production profile as well as other improvements.

PR #587, opened 2016-06-28
Update in HTTP paths with wildcards

PabloCamino:
I have added a note on paths with wildcards after handling a couple cases where overlapping paths lead to an error. The already present note on the most exact path being used does not explain this situation where a request goes unfulfilled.
Available if you have any questions.
Best.


george-hoffer:
Thank you for this update.

PR #590, opened 2016-06-29
Update managing-servers.adoc

suchidesh:
Please link to Cloudhub Networking Guide page where users can get access to how to configure their system using white listed IPs.


nearnshaw:
Suchi, from what I uderstood from talking to Walter, what would need to be linked here is not what's in the CH networking guide (https://docs.mulesoft.com/runtime-manager/cloudhub-networking-guide#ip-ranges-and-static-ips), but rather what's in the Agent installation doc: https://docs.mulesoft.com/mule-agent/v/1.4.0/installing-mule-agent#ips-to-whitelist

PR #591, opened 2016-06-30
typo fix

jmjdamen:
missing 'l' so link was broken / not usable


fermujica:
Thanks for fixing this!

PR #592, opened 2016-07-01
Clarify Mule 3.8 runtime steps

ethanport:
Added notes clarifying Mule 3.8 runtime configuration steps. Specifically that it may be necessary to configure the client id and client secret in the Mule runtime's wrapper.conf file if the runtime is managed by MMC.

### DIFF
--- a/anypoint-platform-for-apis/v/latest/configuring-an-api-gateway.adoc
+++ b/anypoint-platform-for-apis/v/latest/configuring-an-api-gateway.adoc
@@ -1,7 +1,12 @@
 = Configuring an API Gateway
 :keywords: api, cloudhub, gateway, auto-discovery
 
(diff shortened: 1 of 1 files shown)


nearnshaw:
Thanks Ethan, these very valuable clarifications!

krishahn:
Configuring an API Gateway pertains to installations earlier than Mule 3.8.0 as stated in the intro sentences. We are tentatively planning to label this doc ARCHIVE, so putting Mule 3.8.0 info in this doc doesn't seem right. I think what we need is a pointer to this info from Configuring an API Gateway to /mulesoft-docs/mule-fundamentals/v/3.8/deploying-mule-applications.adoc. How about moving the info there? Maybe it's already there and all we need to do is point to it.

krishahn:
Adding "Managing Mule runtime with both MMC and the Anypoint Runtime Manager is _not_ supported." to the overview section of the Runtime Manager page

PR #594, opened 2016-07-01
Update xpath.adoc

tonedef71:
Fixed an itty-bitty typo.


fermujica:
Thanks!! I approved this along with #593

PR #595, opened 2016-07-05
add CI profile to Gradle build

mojavelinux:
- add profile for using Gradle build in CI (-Pprofile=ci)
- load build script per profile
- configure GitHub Pages plugin keep clone between runs
- move all settings into extended project properties
- check for existing assets clone dynamically using an onlyIf block
- don't copy template except when dev profile is active
- revert template after running docs builder when dev profile is active
- fix file property lookup when using single page builder
- update task descriptions


seanosterberg:
Thanks Dan!

mojavelinux:
We're almost ready to put this to use. I still need add the final step, which is to push the changes to the remote build output repository. Then, we need to setup a new Jenkins build and give it a test.

PR #600, opened 2016-07-08
Repaired wrong URLs

MiroslavMeca-Profiq:
Some URLs is wrong in sections _"Installing Tcat Server on Solaris"_ and _"Solaris Operating Systems Supported"_.


george-hoffer:
Thank you Miroslav

PR #603, opened 2016-07-08
Repaired URL for "Symas Corporation"

MiroslavMeca-Profiq:
link http://lirone.symas.net/symas-com/ is wrong, it is return HTTP status 404.
I think that correctly URL can be https://downloads.symas.com


george-hoffer:
Thank you Miroslav!

PR #610, opened 2016-07-09
clarify policy-ordering function

PaulAndersonMule:
policy-ordering button is only active when the policy is being actively managed by a running Gateway

### DIFF
--- a/anypoint-platform-for-apis/v/latest/applying-runtime-policies.adoc
+++ b/anypoint-platform-for-apis/v/latest/applying-runtime-policies.adoc
@@ -46,6 +46,9 @@ The list of any applied policies and available policies appears. The list includ
 +
 . At the top of the *Applied policies* list, click the greyed-out *Edit Policy Order* button to assign the priority.
 +
+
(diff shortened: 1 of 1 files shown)


krishahn:
Thanks for the correction, Paul. The url of the file has changed: https://docs.mulesoft.com/api-manager/using-policies. The file having the old URL is in place today, but will be gone tomorrow. I made the following changes to fold in your correction: 

**Change to the 1st paragraph of Setting the Order of Execution:**

You can set the order of execution of applied policies on an application in Anypoint Platform if you are using one of the following releases:

* Studio 6.0 for creation, deployed to Anypoint Platform with auto-discovery
* Mule 3.8 unified runtime
* API Gateway Runtime 2.2.0

**Change to step 4:**

At the top of the Applied policies list, click the greyed-out (unavailable) Edit Policy Order button to assign the priority.

(screenshot)

   The Edit Policy Order button is available only when the API is actively managed by an API Gateway or Mule 3.8 unified runtime.

PR #611, opened 2016-07-10
Fix broken link:setting up development environment

Nath-Lee:
had "dev" instead of "development" in the url


george-hoffer:
Thank you Nathan

PR #612, opened 2016-07-10
fix broken formatting that was hiding text

Nath-Lee:
Something crazy happening with the formatting/wiki markup, but this displays the currently hidden info at least.

### DIFF
--- a/mule-user-guide/v/3.8/http-request-connector.adoc
+++ b/mule-user-guide/v/3.8/http-request-connector.adoc
@@ -785,7 +785,7 @@ This sends a GET request to link:http://www.example.com/test[www.example.com/tes
 
 == Configuring Source and Target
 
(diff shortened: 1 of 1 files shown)


mojavelinux:
Backticks do not escape the text, so the # symbols are being matched across the line (due to a matching edge case). You have found one of the correct approaches. The more universal approach is to use:

```
`+#[payload]+`
```

Notice the addition of the + symbols, which escapes the inner text. This avoids the need of having to count the # pairs across the line.

More information can be found on the following page: http://asciidoctor.org/docs/migration/#migration-cheatsheet

mojavelinux:
Btw, I've noticed many instances of this particular problem, primarily because the `#[name]` is a common pattern in the docs.

george-hoffer:
Thank you!

PR #617, opened 2016-07-10
link title correction

Nath-Lee:
should have been "setting up an API Proxy"


krishahn:
Thanks for spotting this. Since you issued this pull request, we reorganized the docs and changed this link.

PR #619, opened 2016-07-10
update to 3.8 link for anypoint studio

Nath-Lee:
old version link hanging around


krishahn:
Thanks for flagging this problem. Coincidentially, I noticed and changed it while reorganizing the docs over the weekend, I removed these See Alsos, so I can't merge your pull request. We're changing the Mule Fundamentals sections and links have to be changed and rechanged--better to let readers link from the nav bar, reduce the number of cross-references.

Nath-Lee:
No worries, was just putting things in as I saw them - bound to overlap with some cleanups. :)

PR #621, opened 2016-07-10
formatting cleanup/improvements

Nath-Lee:
had a bunch of dashes in there - complex structure probably needs more of a cleanup but think this helps


george-hoffer:
Thank you.

PR #622, opened 2016-07-10
Bad bookmark anchor for the Precedence Table

cgorshing:
This just corrects two links in the DW to the Precedence Table. I think "Operator Precedence Table" makes more sense as a link, but these were the only two instances that used that name, so I changed them to use "Precedence Table"


nearnshaw:
Thanks Chad! Yes those links needed fixing. The name "precedence table" makes more sense to me than "operator precedence table" because some of those elementsin the table  – such as "when" or "otherwise" –  aren't technically operators, but rather flow control expressions.

PR #629, opened 2016-07-11
Removing "Alerts" section as it's duplicated from its own page

octaviozamudio:
There is no need for this to be in this particular site.

### DIFF
--- a/runtime-manager/v/latest/managing-servers.adoc
+++ b/runtime-manager/v/latest/managing-servers.adoc
@@ -459,16 +459,6 @@ One advantage of using multicast is that a server does not need to be running to
 [NOTE]
 Check with your network administrator if multicast is allowed within your network, as many networks block multicast functionalities.
 
-== Alerts
-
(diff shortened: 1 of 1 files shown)


nearnshaw:
Octavio, I appreciate your feedback, but in this case I disagree. I think it's valuable to have a short mention of the fact that you have the ability to set alerts at a server level in this doc. Traditionally alerts in the platform have all been related to application events, so someone in devops whose role is to deal with servers might not even think of looking in the alerts page. These alerts in particular are specially relevant to that person, and we need to at least link to the alerts page from here.
I'll reject this pull request. On the other hand, I will do edits on the page to make this section shorter and to change the note that says that alerts are only set for individual users, as you changed on your other pull request.
Thanks!
Nico

octaviozamudio:
Nicolas,
Okay! LGTM as long as that note regarding the disabling of alerts is rectified.

Thanks to you for the review!!

PR #630, opened 2016-07-11
Fixed to match the current behavior

octaviozamudio:
Changed the [NOTE] in "Switching an Alert Off" so it matches the product's real behavior.


nearnshaw:
Thanks Octavio! I remember hearing alerts would be switched on/off for individual users, but I think that was at a point where there was only a spec for the feature... I trust that what you say is how the real feature actually works :)

PR #631, opened 2016-07-11
Removed spacing which caused exceptions for 3.8.0

cgorshing:
In using http:success-code-validator I used the format as mentioned in the docs: "200, 404" and I received an error of: Caused by: java.lang.NumberFormatException: For input string: " 404"

So this PR changes the docs to use the format: "200,404" (not the space has been removed).

I'm not sure if this is wrong in the docs, or wrong in the code.


george-hoffer:
Thank you, Chad, and thank you for fixing all 3 versions!!

PR #633, opened 2016-07-12
Clarification around workday as a dependency

cgorshing:
When I was reading https://docs.mulesoft.com/mule-user-guide/v/3.8/workday-connector I was pretty confused on the "Upgrading from an Older Version" table. First I thought there were two different paths to upgrading, depending upon what you are using now. Then the artifact element used angle brackets as well to denote, what I think, a square bracket should be used for.

So I just removed both entries as the outcome is the same, just listed out the current way to denote WD as a dependency.

The only thing I could think of is calling out that there are no longer any WD modules.


nearnshaw:
Thanks Chad, that makes a lot of sense to me too, I'm merging your changes!

PR #637, opened 2016-07-13
Update using-dataweave-in-studio.adoc

tonedef71:
The <dw:input-payload> element may only appear once within an enclosing <dw:transform-message> element; it may not be repeated.

Cleaned up some whitespace indents.


nearnshaw:
You're absolutely right!  I don't know how those examples ended up looking like that but thanks for fixing them Tony! I'll merge these changes and do the same fixes on 3.7

PR #638, opened 2016-07-13
Include tip for the FTP configuration

WilliamPinho:
The image suggests that no user nor password should be set up. However, it is mandatory to set up the credentials in order use the FTP endpoint connector.


fermujica:
Thanks! I added your contribution to our docs.

PR #647, opened 2016-07-19
Move less compilation from client to the gradle build

hsablonniere:
For now it's just a *work in progress* but first metrics indicate good perf improvements...

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,9 @@
 // TODO
 // * add mode to run jetty in background
-// * combine css (need to update template as well)
 // * set outputs.dir for custom tasks
+import com.eriwen.gradle.css.tasks.CombineCssTask
+import com.eriwen.gradle.css.tasks.MinifyCssTask
+
 buildscript {
   configurations.create('builder')
   configurations.create('singlePageBuilder')
@@ -137,7 +139,7 @@ task copyAssets(type: Copy) {
   from(assetsDir) {
     exclude 'README.md'
     exclude '_*/**'
-    //exclude 'css/*.less'
+    exclude 'css'
   }
   into siteDir
   //outputs.dir siteDir
@@ -152,10 +154,15 @@ task copyAssets(type: Copy) {
 
 css {
   source {
-    main {
+    standard {
       css {
         srcDir "$assetsDir/css"
         include '*.css'
+      }
+    }
+    less {
+      css {
+        srcDir "$assetsDir/css"
         include '*.less'
       }
     }
@@ -165,20 +172,36 @@ css {
 lesscss {
   dependsOn setup
   mustRunAfter copyAssets
-  source = css.source.main.css.asFileTree
-  dest = file("$siteDir/css/")
+  source = css.source.less.css.asFileTree
+  dest = "${buildDir}/generated-css/mulesoft"
 }
 
-//def cssDir = "app/styles/"
-//def cssSrc = ["file1.css","file2.css"]
-//
-//combineCss {
-//    source = cssSrc.collect {cssDir+it}
-//    dest = "$buildDir/all.css"
-//}
+task combineMulesoftCss(type: CombineCssTask, group: 'Build') {
+  source = lesscss
+  dest = "${buildDir}/css/mulesoft.css"
+}
+
+task combineVendorCss(type: CombineCssTask, group: 'Build') {
+  source = css.source.standard.css.asFileTree
+  dest = "${buildDir}/css/vendor.css"
+}
+
+task minifyMulesoftCss(type: MinifyCssTask, group: 'Build') {
+  source = combineMulesoftCss
+  dest = "${siteDir}/css/mulesoft.min.css"
+}
+
+task minifyVendorCss(type: MinifyCssTask, group: 'Build') {
+  source = combineVendorCss
+  dest = "${siteDir}/css/vendor.min.css"
+}
+
+task minifyAllCss(group: 'Build') {
+  dependsOn minifyMulesoftCss, minifyVendorCss
+}
 
 task build(group: 'Build', description: 'Builds site.') {
-  dependsOn setup, buildHtml, copyAssets //, lesscss
+  dependsOn setup, buildHtml, copyAssets, minifyAllCss
   mustRunAfter clean
 }
 


build.gradle
```diff
@@ -127,41 +128,56 @@ task copyAssets(type: Copy) {
...
+    exclude 'css'
   }
   into siteDir
   //outputs.dir siteDir
 }
 
 css {
   source {
-    main {
+    assets {
```
mojavelinux: Perhaps call this block "static" instead of "assets". That pairs better with "less".

build.gradle
```diff
@@ -127,41 +128,56 @@ task copyAssets(type: Copy) {
...
+  source = css.source.less.css.asFileTree
+  dest = file("$buildDir/mulesoft-css/")
+}
+
+task combineMulesoftCss(type: CombineCssTask, group: 'Build') {
+  source = lesscss
+  dest = "${buildDir}/css/mulesoft.css"
+}
+
+task combineVendorsCss(type: CombineCssTask, group: 'Build') {
```
mojavelinux: Let's use the singular "vendor" instead of "vendors". That's the correct form for the adjective, I'm pretty sure.

build.gradle
```diff
@@ -127,41 +128,56 @@ task copyAssets(type: Copy) {
...
+        srcDir "$assetsDir/css"
         include '*.less'
       }
     }
   }
 }
 
 lesscss {
-  dependsOn setup
-  mustRunAfter copyAssets
```
mojavelinux: You'll still need lesscss to depend on setup (which clones the assets repo). The mustRunAfter is merely a personal preference as to when you want the task to run after setup.

The other task dependencies (combine and minify) are calculated automatically based on the fact that the task inputs are wired to the outputs of the other tasks.

build.gradle
```diff
@@ -127,41 +128,56 @@ task copyAssets(type: Copy) {
...
-  source = css.source.main.css.asFileTree
-  dest = file("$siteDir/css/")
+  source = css.source.less.css.asFileTree
+  dest = file("$buildDir/mulesoft-css/")
```
mojavelinux: I think this folder should be:

```
file("$buildDir/generated-css/mulesoft/")
```

hsablonniere:
I applied the correction in a [fixup commit](https://robots.thoughtbot.com/autosquashing-git-commits) ;-)

build.gradle
```diff
@@ -5,6 +5,7 @@
 import com.palominolabs.gradle.task.git.clone.GitCloneTask
 import org.eclipse.jgit.api.Git
 import org.eclipse.jgit.lib.RepositoryBuilder
+import com.eriwen.gradle.css.tasks.*
```
mojavelinux: Could you resolve this import to specific classes for clarity? (sorry I missed this the first time)

build.gradle
```diff
@@ -127,18 +129,23 @@ task copyAssets(type: Copy) {
...
+    exclude 'css'
   }
   into siteDir
   //outputs.dir siteDir
 }
 
 css {
   source {
-    main {
+    staticAssets {
```
mojavelinux: I understand we can't use `static`. How about `native`, `plain` or `normal` instead? I'm leaning more towards `native` if it's allowed.

build.gradle
```diff
@@ -48,7 +50,7 @@ group = 'com.mulesoft.documentation'
 ext {
   assetsDir = '_assets'
   assetsRepoBranchName = 'staging'
-  assetsRepoUri = 'https://github.com/mulesoft/mulesoft-docs-site-assets'
+  assetsRepoUri = 'https://github.com/hsablonniere/mulesoft-docs-site-assets'
```
mojavelinux: Can you revert the repository URL change in the PR?

We'll need to sort out how to allow a different URL to be used locally for development. The simplest way is to clone the repo to _assets yourself and avoid running the clean task. As I mentioned in our chat, I'll add something like a check for `.git-keep` to prevent clean from removing the cloned folder.

build.gradle
```diff
@@ -148,20 +155,32 @@ css {
...
+}
+
+task minifyVendorCss(type: MinifyCssTask, group: 'Build') {
+  source = combineVendorCss
+  dest = "${siteDir}/css/vendor-min.css"
+}
 
 task build(group: 'Build', description: 'Builds site.') {
-  dependsOn setup, buildHtml, copyAssets //, lesscss
+  dependsOn setup, buildHtml, copyAssets, minifyMulesoftCss, minifyVendorCss
```
mojavelinux: I think we should add a new task named `minifyCss` that depends on `minifyMulesoftCss` and `minifyVendorCss`. That way, we can abstract away the details a bit.

hsablonniere: I had one at first but it felt redundant. I'll add it back.

build.gradle
hsablonniere: This is done is a temp commit https://github.com/mulesoft/mulesoft-docs/pull/647/commits/93c3d5fcbe0122a2a1e4c02052a0e9160acf61b5 that must be removed at rebase.

build.gradle
hsablonniere: `native` is not allowed so I used `basic`.

build.gradle
hsablonniere: I named it `minifyAllCss` since `minifyCss` is already defined by the css plugin.

hsablonniere:
@mojavelinux Latest comments applied and branch rebased on latest `mulesoft/master`.

mojavelinux:
Très bien !

Now we just need to coordinate the merge. I'll work with @seanosterberg on this because it has to be performed in sync with the assets repo.

mojavelinux:
When comparing the site with the compiled LESS against the site with the raw LESS, I notice some elements are shifting. This is likely due to an ordering issue in the CSS. When combining the LESS files, we need to preserve the current order (as seen in the template file).

hsablonniere:
It seems I need to explicitely declare the files. Do you agree with this?

https://github.com/eriwen/gradle-css-plugin/issues/11

hsablonniere:
You mean with numbers like `01-zzz.css`, `02-vvv.css`, `03-aaa.css`... ?

mojavelinux:
Yes, or any other prefix that would allow sorting. The more I think about it, the more I think this is the right approach. Not only is it extensible (without having to maintain the build), it also communicates the order to the CSS author without having to look in the build. Win win!

build.gradle
```diff
@@ -127,18 +128,23 @@ task copyAssets(type: Copy) {
...
+    exclude 'css'
   }
   into siteDir
   //outputs.dir siteDir
 }
 
 css {
   source {
-    main {
+    basic {
```
mojavelinux: After giving this more thought, I'd like to change "basic" to "standard". I think the term "standard" fits well here because it is "standard CSS" or the "CSS standard syntax".

mojavelinux:
We are still getting differences because, despite what we had expected, Gradle does not pick up the files in the order they are sorted on the filesystem. Therefore, we need to explicitly sort them. I'll add code review comments on the lines where this needs to happen.

I'd also like to suggest renaming 01-asciidoctor.css to 01-asciidoctor.min.css (since it's a minified file). If we also replace 03-jquery.fancybox.css with the minified version, then we don't need to minify the vendor CSS (just combine it). I'd like to make that switch.

build.gradle
```diff
@@ -165,20 +172,36 @@ css {
...
 
-//def cssDir = "app/styles/"
-//def cssSrc = ["file1.css","file2.css"]
-//
-//combineCss {
-//    source = cssSrc.collect {cssDir+it}
-//    dest = "$buildDir/all.css"
-//}
+task combineVendorCss(type: CombineCssTask, group: 'Build') {
+  source = css.source.standard.css.asFileTree
```
mojavelinux: This the following assignment instead to sort the standard CSS files:

```
source = files(css.source.standard.css.files.sort())
```

build.gradle
```diff
@@ -165,20 +172,36 @@ css {
...
   mustRunAfter copyAssets
-  source = css.source.main.css.asFileTree
-  dest = file("$siteDir/css/")
+  source = css.source.less.css.asFileTree
+  dest = "${buildDir}/generated-css/mulesoft"
+}
+
+task combineMulesoftCss(type: CombineCssTask, group: 'Build') {
+  source = lesscss
+  dest = "${buildDir}/css/mulesoft.css"
```
mojavelinux: Add the following block after the dest assignment to sort the mulesoft CSS:

```
  doFirst {
    source = files(source.files.sort())
  }
```

(I can't figure out any other way to do it).

mojavelinux:
In both minify CSS tasks (or just the mulesoft CSS if we drop minimization of the vendor CSS), add the following configuration block to set the maximum line width:

```
  yuicompressor {
    lineBreakPos = 500
  }
```

This is what the Google Closure compiler team recommends.

mojavelinux:
This pull request makes me realize that we're going to need to use an explicit tag / commit of the assets even in dev and staging. Otherwise, long-running docs branches will get the new build but not the new template & css. I'll make that change and, at the very least, update the stage branch in this repository so it produces good output on http://docs-stg.mulesoft.com/.

mojavelinux:
The assets are now pinned in all build profiles.

mojavelinux:
Instead of renaming 01-asciidoctor.css to 01-asciidoctor.min.css, I decided to switch to the unminified version of bootstrap and allow yuicompressor to minimize it all.

I applied the recommended fixes to get the ordering right. I also slightly reworked the wiring of the tasks so that we don't get bitten by Gradle's up-to-date check.

mojavelinux:
The changes from this PR have been moved to #829.

PR #649, opened 2016-07-19
Fix tab syntax

mojavelinux:
- fix invalid tab blocks
  - fix cases where "tabs" block style is applied to an individual tab
  - fix missing delimiters around tab blocks
  - add missing closing delimiter for tab container
  - remove blank lines between block attribute list and block delimiter

### DIFF
--- a/mule-healthcare-toolkit/v/3.4/hl7-exception-strategy.adoc
+++ b/mule-healthcare-toolkit/v/3.4/hl7-exception-strategy.adoc
@@ -27,7 +27,6 @@ hl7:exception-strategy
 [tabs]
 ------
 [tab,title="Studio"]
-
 ....
 [source, xml, linenums]
 ----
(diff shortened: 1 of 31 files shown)


mojavelinux:
@seanosterberg This will need to be merged before we can make the switch to the new tab extension.

mojavelinux:
This is causing an unexpected failure in the build. I'm going to need to find a way around it.

mojavelinux:
Okay, I'm going to have to hold off on adding the discrete block style until we switch to the new tabs extension. The section navigator is looking for headings inside the tab content and failing when it doesn't find a link. We'll just make the switch to the new docs builder and the addition of the discrete block style at the same time.

I'll update this pull request just to include the first commit, which fixes some broken tab syntax.

mojavelinux:
I want to emphasize that no time was lost since I have a script that is making this change.

mojavelinux:
@seanosterberg This pull request is ready to merge. It merely fixes broken/invalid syntax that corrects the HTML structure on pages such as https://docs.mulesoft.com/mule-user-guide/v/3.4/installing#which-version-of-mule-am-i-using and https://docs.mulesoft.com/munit/v/1.1.1/example-testing-apikit#the-munit-test.

PR #650, opened 2016-07-19
Update links in mule-object-stores.adoc

jfielding:
Updating the links to correctly reference the cache scope


george-hoffer:
Thank you Jay

PR #651, opened 2016-07-20
remove misplaced AsciiDoc files

mojavelinux:
These files are unused and are being copied directly to the output. You can see from the URL below that the raw AsciiDoc can be accessed directly from the production site:

https://docs.mulesoft.com/ule-user-guide/v/3.2/_images/adding-business-logic-to-a-flow.adoc


mojavelinux:
@seanosterberg I think this can be merged immediately. I don't know how these AsciiDoc files ended up in the _images directory, but I'm confident they are not being used. They keep getting in the way when searching for syntax to fix in the AsciiDoc files.

PR #656, opened 2016-07-21
CORS and mixed content warning issues should be addressed properly.

fermujica:
When accessing an API behind a firewall, the CORS settings need to be well configured. 
We should not recommend users to disable their browser´s security settings unless they are using a controlled environment, and such practice should be used only for testing (never for production).


krishahn:
tks, Fer!

PR #657, opened 2016-07-21
Fix Studio Update site URL for version 1.5.1

anosenzo:


fermujica:
Thanks for fixing that!

PR #659, opened 2016-07-22
Change Mule ESB Server to Mule runtime

ethanport:
The product was re-branded as Mule runtime


ethanport:
Same probably applies to the 3.7 doc.

george-hoffer:
Thanks, Ethan

PR #660, opened 2016-07-22
Update configuring-maven-to-work-with-mule-esb.adoc

adrianmartinmulesoft:
Added the public repository reference. Several cases were fixed by adding this repo.


george-hoffer:
Thank you Adrian

PR #661, opened 2016-07-22
Reworded Dynamic Keys explanation

cgorshing:
The docs had:

> In order to set an a key via an expression

And I didn't feel like that made much sense, so I changed it to:
> In order to specify a key via an expression


george-hoffer:
Thank you Chad

PR #669, opened 2016-07-25
Update mule-expression-language-reference.adoc

tonedef71:
Reversed order of words so that the sentence makes sense.


george-hoffer:
Thank you, Anthony!

PR #670, opened 2016-07-26
Update dataweave-operators.adoc

adrianmartinmulesoft:
Correct an error un filter() specs for object. The key is an index and it's documented as key


nearnshaw:
Thanks Adrian!

PR #671, opened 2016-07-27
Version is incorrect

adrianpineyro:
Please update page name as well.


nearnshaw:
Thanks Adri, I'll change the page URL as well

PR #674, opened 2016-07-28
typo in keyalg option

jmjdamen:
typo fix


fermujica:
Thanks for fixing this!!

PR #677, opened 2016-07-29
Repaired wrong URL for "MISSING readme"

MiroslavMeca-Profiq:
I think I found correct link https://github.com/mulesoft-labs/rajapa/blob/5a558ee98236991aaf49d3d03b83b62c48dd1c86/MISSING.md
Previous URL returns HTTP status code 404, Page not found

### DIFF
--- a/release-notes/v/latest/raml-1-early-access-support.adoc
+++ b/release-notes/v/latest/raml-1-early-access-support.adoc
@@ -277,7 +277,7 @@ types:
 * In JavaScript parser (i.e. API Designer, API Console): VALID
 * In Java parser (i.e. RAML editor in Studio, APIkit): INVALID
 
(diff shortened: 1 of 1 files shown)


krishahn:
Thanks for flagging the problem The link you provided goes to the same info as the "RAML 1.0 features gap" https://github.com/raml-org/raml-java-parser/blob/0.1.0-alpha-2/MISSING.md, which we mention at the end of the [Java Parser Known Gaps](https://docs.mulesoft.com/release-notes/raml-1-early-access-support#java-parser-known-gaps) section. I think the link at the bottom of the page (the link that your PR is fixing) went to a more complete list of Java parser problems than the gaps between the parser and the RAML 1.0 early access release. Why link to the same information twice, right? I'll check with the RAML guru in London to see where the "more complete" list was moved. In the meantime, I'll remove the broken link.

PR #678, opened 2016-07-29
Update app domain name length

nalnaji:
Updated incorrect domain name length for cloudhub application


nearnshaw:
Thanks Nalnaji, you are right about that!

For future contributions, we are required to ask you to please take a moment to sign our contributor's agreement as described at https://github.com/mulesoft/mulesoft-docs/blob/master/README.adoc . That way, you'll be registered as a contributor and we can accept more of your pull requests in the future.

PR #685, opened 2016-08-01
It's after 30 July

ethanport:
Removed pre-release note, as it has now been released.


fermujica:
Thanks Ethan!
Yes, I was about to remove that as well.

PR #687, opened 2016-08-03
url markdown fix

jmjdamen:
could not test with relative path please check


nearnshaw:
Thanks Joost!  I think your fix of the link is what makes the most sense, that extra part of the address wasn't really adding anything

PR #688, opened 2016-08-03
added OpenAM + important markdown

jmjdamen:
OpenAM url + Importance markdown for older versions


nearnshaw:
Worth adding, thanks!

PR #693, opened 2016-08-03
clarified -S option

PaulAndersonMule:
@suchidesh @ sramirezayuso I think this clarification does it. Work for you?

docs-team please defer until suchidesh & or sramirezayuso confirm it's good.

original Slack msg here https://mulesoft.slack.com/archives/anypoint-runtime-mgr/p1470250370000257


nearnshaw:
Thanks Paul, Santi gave this a look and agrees with your comment, so I'll go ahead and merge this change!

PR #698, opened 2016-08-09
Update Resetting Your Password

cmordue:
Corrected the instructions for resetting your password and forgot username.


krishahn:
Thanks.

PR #699, opened 2016-08-09

adrianpineyro:
Clarification on message's properties when using the lookup function.


nearnshaw:
Thanks Adri, very valuable info

PR #702, opened 2016-08-10
Correct API version

NathanNam:
Update Workday API version to 26 and Devkit version to 3.8.2


nearnshaw:
Thanks Nathan, merging changes now!

PR #703, opened 2016-08-10

tonedef71:
Removed redundant, misplaced "Warning" block.  The same warning some lines up at the beginning of the "Input Directive" section should be enough warning for readers of this page.


nearnshaw:
Thanks Tony! Yeah, agree that although this is an important warning, this repetition is too redundant

PR #704, opened 2016-08-10

tonedef71:
A couple of minor punctuation edits.

Removed the "tip" concerning adding an %input directive outside of Studio because it seems to be made moot by other statements.

Re-organized and consolidated information about the %input directive.  Also added some information taken from the 3.7 documentation.

Information about the %input directive should be maintained in the documentation until it is no longer valid.  By the way, if the %input directive is slated for deprecation in a future release, it should also be described in this documentation as such.

Please feel free to amend this proposed documentation change as necessary.


nearnshaw:
Thanks Tony! These changes make a lot of sense to me :)

PR #707, opened 2016-08-11
update template to sync with site assets update

mojavelinux:


mojavelinux:
This PR is no longer needed now that the templates live in the assets repository.

PR #709, opened 2016-08-13
clarified exception rules

ethanport:


george-hoffer:
Thanks Ethan. I also fixed in previous mule doc versions.

PR #714, opened 2016-08-14
add AP dev portal

ethanport:
The anypoint platform developer portal is difficult to find with a simple Google search. We should add this link in other parts of the doc.


krishahn:
OK, adding the link to https://docs.mulesoft.com/api-manager/engaging-users-of-your-api and https://docs.mulesoft.com/api-manager/managing-your-api (w/b published in 30 min or so).

PR #715, opened 2016-08-15
Updated name of variable app.workDir

MichalKalita-profiq:
app.workdir is null
app.workDir is work directory of application


george-hoffer:
Thanks for helping improve MuleSoft documentation

PR #725, opened 2016-08-18
The file extension must be in lowercase.

angelspano:
We had cases where if the application has .ZIP extension (uppercase) isn't  being deployed.

### DIFF
--- a/mule-user-guide/v/3.8/application-format.adoc
+++ b/mule-user-guide/v/3.8/application-format.adoc
@@ -3,7 +3,7 @@
 
 The deployment unit of a Mule application encapsulates everything an application needs to function, such as libraries, custom code, configuration, deployment descriptor and any environment properties accompanying the application. It exists in one of two formats:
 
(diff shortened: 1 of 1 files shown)


fermujica:
Hi Angel, thanks for pointing this up.
It's a fair warning, and I think it makes sense to keep your file extensions lowercase. 
I've been uploading zip projects with the uppercase file extension and they have been working. 

In order to properly document this, I'd like to have a little more context.
Do you have a specific use case when a customer got an error? Did they have a specific error description? 
In which scenario would a customer have to manually modify the .zip file extension to uppercase?

angelspano:
Hi Fer, 
Yes we have a case in which the customer has automated some post-processing on the mule artifact and their zip files are renamed to uppercase. 
I've replicated the situation in my machine and I haven't been able to deploy a zip file with the extension in uppercase. I've checked the source code and found out that the string ".zip" is hardcoded  and it's used to filter files to be deployed and to create the expanded folders. 
I've also created an enhancement request to fix this.

PR #727, opened 2016-08-19
Corrected prompt for interactive mode commands

adobni:


fermujica:
Thanks Alejandro. 
 Indeed looks better to have the prompt character in the example reflecting how the real CLI looks like.

I'm merging your PR.

PR #728, opened 2016-08-19
Update mule-agent.adoc

suchidesh:
...and so on and so forth...there's many more but I'm sure it's just easy to do a find+replace. Also it needs to change on the TOC on the side.


nearnshaw:
Thanks Suchi, next week I'll take care of the other uses of the term and changing it on the TOC too

For now I'll merge the changes here

PR #733, opened 2016-08-23
rewrite prod build profile to use native git commands

mojavelinux:


mojavelinux:
I've tested this new production profile locally and using the docs-prod-GRADLE CI job and I'm now confident that it works as expected. @sharkberto can I merge? Note that merging this right now does not impact the production site at all. It just puts us a step closer to using it.

mojavelinux:
@sharkberto Can I merge this PR?

PR #738, opened 2016-08-23
Update munit-maven-support.adoc

diegostrubolini:
Closed **note** section because if not, all the text after it was inserted inside


fermujica:
Oops. Thanks for fixing that. I don't know how that broke to begin with.

PR #739, opened 2016-08-23
Renamed Mule agent

ethanport:
Runtime Manager Mule agent is helpful to avoid confusion with the older MMC Mule agent docs. Also clarified details about amc_setup script options and the effects of the script.


nearnshaw:
Thanks Ethan! Although I believe this is a valuable change, the naming we agreed on with Suchi was "Runtime Manager Agent", rather than the "Runtime Manager Mule Agent". So I won't be merging this pull request, but I will be implementing this change throughout the entire docs site very soon!

PR #741, opened 2016-08-24
How external identity applies in Business Groups

adobni:


fermujica:
Thanks Alejandro. 
I moved this consideration to the top of the page as it seems like a fair warning to give users earlier in the doc.

See [f27b9b7](https://github.com/mulesoft/mulesoft-docs/commit/f27b9b7f8af819a98de940b51e8e4cb104731d49)

PR #745, opened 2016-08-26
fix staging profile of Gradle build

mojavelinux:
- configure robots.txt to disable crawling
- only preserve file timestamps when transfering files to staging server (addresses task failure)


mojavelinux:
@sharkberto This patch is needed to allow us to switch the docs-staging CI job to use the Gradle-based build.

PR #750, opened 2016-08-26
Munit terms

krishahn:
In sweeping the docs for gateway/proxy terms to clarify for University of Austin, these domain references seemed a little inconsistent with the terms used in the main domain docs.

### DIFF
--- a/access-management/v/latest/organization.adoc
+++ b/access-management/v/latest/organization.adoc
@@ -7,7 +7,7 @@ As an <<The Organization Administrator Role,administrator of an Organization>> i
 * Assign users to roles to define their permissions in the platform
 * Edit and remove users from your organization
 * Configure organization settings
-* Access your organization's client ID and client secret, needed for configuring your API Gateway instances
+* Access your organization's link:/access-management/organization#client-id-and-client-secret[client ID and client secret]
 * Access analytics for the APIs in your organization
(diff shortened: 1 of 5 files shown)


access-management/v/latest/organization.adoc
```diff
@@ -43,7 +43,7 @@ image::organization-11dbd.png[organization-11dbd]
 
 Even though multiple organizations can be created by different users using the same company name, each organization has a unique domain.
 
-Here you can also consult your organization's Client Id and Client Secret. These are useful if you want to use an API Gateway, as you must register it with a specific organization. The Client Id and Client Secret that belong to the Master Organization provide all of the permissions from the Business Groups within it.
+Here you can also consult your organization's Client Id and Client Secret. These are useful if you want to use an API Manager, as you must register it with a specific organization. The Client Id and Client Secret that belong to the Master Organization provide all of the permissions from the Business Groups within it.
```
fermujica: How  is API Manager using your organization's Id and secret ?

krishahn:
API Manager, which replaced the API Gateway term in most cases (per Eva), when the unification with Mule occurred in 3.8.0, uses the client Id and client secret in these ways: 

* policy enforcement
* applying custom policies to avoid clashes between policies
* building an OAuth2 provider
* testing the provider
* in the developer portal, which is accessed through API manager and documented in the API Manager section
* registering an API in API Manager
* checking if a client calling an API registered with API manager has been given access to it

The docs talk about configuring API Gateway in the archive section instead of in the API Manager section, because generally, in 3.8.0 and later, no configuration is needed.

PR #752, opened 2016-08-27
Update runtime-manager-1.5.1-release-notes.adoc

suchidesh:
I would remove the on-prem version tip, at least from the release notes. It's confusing because it makes it seem like they are the same versions and they are not.

Thanks,
Suchi


nearnshaw:
Yup, completely agree!
This comment was copied from a couple of older release notes, I'll remove it from there as well.

PR #754, opened 2016-08-29
Update installing-an-enterprise-license.adoc

adrianpineyro:
Adding clarifications: Mule Server must not be running


nearnshaw:
Thanks Adri! Merging this change now

PR #759, opened 2016-08-31
sub-org > business group

krishahn:
Please see email from A. McB., need to fix before responding.


fermujica:
This is great!
Yes, the sub-org is an old term that keeps being used during development and sometimes finds its way to the docs.
I'll start sanitizing these occurrences. 

Thanks again!

PR #760, opened 2016-08-31
Update Persisting MMC Data to MS SQL Server

jvkaam:
- Line 79 The example for the telnet action was empty (added that)
- Line 123 The download link for the sqljdbc4.jar at Microsoft is correct but the actual download gives a Page not found.
- Line 144 At topic 'Unpacking gzipped tar files' misses the actual example for the command (added that)
- Line 216 'env.dbschema' should be MMC_STATUS (changed that)
- Line 282 'mmc.tracking.db.dbname' should be MMC_PERSISTENCY (changed that)


george-hoffer:

george-hoffer:
Hi John - Thank you for accepting the contributor's agreement and for fixing these issues in the MuleSoft docs.

PR #762, opened 2016-08-31
Document how to build a single page using the Gradle build

mojavelinux:


mojavelinux:
@sharkberto I don't see any reason why this can't be merged right away. May I proceed?

PR #767, opened 2016-09-02
Update until-successful-scope.adoc

angelspano:
We had a case where the customer was using the same object store in cloudhub to store Until Successful data and also to store some key as watermarks. After reviewing the until successful scope, we concluded that if you use the same object store in a until successful scope and in a object store connector, the until successful will delete the entries on application redeploy.


george-hoffer:
Thank you Angel

PR #778, opened 2016-09-08
Update set-message-processor.adoc

diegostrubolini:


fermujica:
Thanks! 
I'm also replicating this chage to older versions: [53f0885](https://github.com/mulesoft/mulesoft-docs/commit/53f08856b6d1d8aa383c431661c7092cd65f606c)

PR #788, opened 2016-09-13
Remove application props/add object store details

ethanport:
Application properties are not related to object stores. The link is called "Application Data", but it shows key-value pairs for the default user object store. It does not show Properties. Also added some other details.


ethanport:
Should also add a correct screenshot of the Application Data tab.

PR #797, opened 2016-09-15
fixed TOC breadcrumb for installing runtime manager agent

nuggetman:
The current docs page lost the TOC breadcrumb, i.e. https://docs.mulesoft.com/runtime-manager/installing-and-configuring-runtime-manager-agent


nearnshaw:
Thanks Alex!
You're right, that link was incorrect. Fortunately there was a redirect in place to prevent people from landing on a 404, but for several reasons it's always better to have the correct links in place :)

PR #798, opened 2016-09-15
stage test

IanWhyte:
In order to create this test, I 
Checked out the stage branch of mulesoft-docs upstream
Made and saved a change
Committed and pushed 
Pushed my copy of stage to origin
Created a pull request

### DIFF
--- a/anypoint-b2b/v/latest/document-types.adoc
+++ b/anypoint-b2b/v/latest/document-types.adoc
@@ -3,7 +3,7 @@
 
 == Add a new document
 
-Click *New*. The Document page appears.
+Click *New*; The Document page appears.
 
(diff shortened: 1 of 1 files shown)


IrfanBaqui:
This works

IrfanBaqui:
But it'll be difficult to maintain two copies. You will have to force push the latest version of mulesoft:stage to modusintegration:stage before making changes just to make sure you're working on the latest copy. We should give you fork permissions...

IanWhyte:
Agreed, forking would be easier. However, doesn't it put us in the position of having to have someone from mule process a pull request anytime I want to stage something?

Ian

-----Original Message-----

PR #799, opened 2016-09-15
Update creating-an-odata-api.adoc

mcoaker:


krishahn:
@mcoaker Thanks for the corrections. Please review the APIkit OData Extension Reference, which is on a separate page, especially this section: https://docs.mulesoft.com/apikit/apikit-odata-extension-reference#implementing-endpoints

PR #800, opened 2016-09-15
Formatting, table fixes

IanWhyte:
Includes fixes that enable tables to display properly.


anypoint-b2b/v/latest/document-types.adoc
```diff
@@ -447,14 +451,14 @@ None
 == Format (JSON)
 
 [%header,cols="3s,7a"]
+
 |===
```
mojavelinux: Please don't include blank lines between the block attributes and the opening table delimiter.

```
[%header,cols="3s,7a"]
|===
```

instead of

```
[%header,cols="3s,7a"]

|===
```

Thanks!

IanWhyte:
Thanks for the input, Dan.

I tend to include a lot of blank lines in my code - it helps me see things better. Are there any other places you can think of where, in an .adoc file, that's an issue?

Ian

-----Original Message-----

mojavelinux:
In many cases, blank lines have meaning, so you have to be careful where you use them. The general rule of thumb is as follows:

* no blank line between a block and its block attributes
* one blank line before and after a section title (where section title includes the block attributes)
* one blank line between paragraphs and delimited blocks
* blank line between table rows, but not between cells in the same row*
  - I also recommend putting each table cell on its own line, except for cells in the header row
* no blank lines within a list item that has complex content (i.e., a list item that uses list continuations)

\* Keep in mind that a blank line after the first row in a table makes that row a header row

If you have a question about a specific case, don't hesitate to ask (dan [at] opendevise [dot] io).

IanWhyte:
Thanks, Dan. I've fixed that occurrence of this issue. Is the next step yours?

IanWhyte:
Thanks, Dan, that's useful information.

About the mechanics of working with PR 800: I've fixed the issue and added
a comment indicating I've done so. Will you see the comment if you're not
actually looking for it? Meaning, what are the appropriate steps for
addressing comments in a PR on the way to accepting it?

Thanks, Ian

mojavelinux:
> I've fixed that occurrence of this issue. Is the next step yours?

As I'm a general adviser, you don't need sign off from me. I'm just offering suggestions / guidelines that I think are going to help keep the source tidy and consistent. As I can ask is that you incorporate those suggestions, as you have done. Thanks!

george-hoffer:
Thanks Ian

PR #806, opened 2016-09-16
Update using-policies.adoc

evangelina:
The default order of policies only applies to Gateways with version 2.2.0 and above so I moved the default order section to be part of the section that describes how to set the order. I don't know if the wording is the most accurate, please feel free to adjust it.


krishahn:
Thanks, Eva.

PR #816, opened 2016-09-22
use master HEAD from mulesoft-docs-site-assets instead of staging HEAD

hsablonniere:


mojavelinux:
Can you update this PR (or add another PR) to point the prod build at the latest commit, which is `b0d8278237ff8f66f3c56f8ac5e91329c75f299f`? In other words, update the sha1 hash in gradle/prod-profile.gradle.

PR #819, opened 2016-09-22

nearnshaw:
The section was named "deploy a proxy to CloudHub" but it actually covered both the use case of deploying CloudHub and to an on-prem server...  this made it confusing, and is related to what the people in support said about distinguishing Runtime Manager from CloudHub. I hence called it "Deploy a proxy" and added sub-sections for each use case.

I also added a new third sub-section below this for explaining a third proxy deployment modality that will appear with PCF. This section is intentionally commented out, as it's a work in progress and talks about something that will only be available with the PCF release.

I also raised the priority of "Deploying a proxy" and "Downloading a proxy" from level 3 to level 2 so that they have more visibility.


krishahn:
Merging, but now Deploying a Proxy to API Gateway should not be above Deploying a Proxy, so I'll move it below Deploying a Proxy and demote its heading level.

PR #821, opened 2016-09-23
Update http-static-resource-handler.adoc

adrianmartinmulesoft:
Clarification for mavenized projects because of the building process


george-hoffer:
Thank you Adrian

PR #826, opened 2016-09-27
Clarifying the ways to expose the API Console

FedeAmdan:


krishahn:
Thanks for reviewing and clarifying things.

PR #828, opened 2016-09-27
section on how to create API versions

nearnshaw:
Added a section to the doc about creating APIs
This is just a draft, feel free to change anything here. I put a placeholder for a screenshot which doesn't exist.

I witnessed how a dev here nearly lost his mind wondering how to create a new version for an existing API. It turned out that his user didn't have permissions for that task, which is odd given that he could create brand new APIs if he wished.  It took us a while to figure that out, and the docs weren't of help for this.

I figured the docs could use a note about this, but when I looked I noticed that there was no appropiate place for that note. So I went ahead and also created a section about the act of creating versions, which I think is also valuable to have.


nearnshaw:
On that note, it's worth clarifying the distinction between the roles *API version owner*, and *API creator*.  Someone may have complete access to editing an API version, but that doesn't entitle him to create a new version.  A link to https://docs.mulesoft.com/access-management/roles would be useful.

krishahn:
Thanks, Nico. There already is a section ["Creating a New Version of an API"](https://docs.mulesoft.com/api-manager/managing-api-versions#creating-a-new-version-of-an-api). I'll fold info in your contributions into the existing section and put a pointer to it from the "Creating an API" section.

PR #830, opened 2016-09-28
relnotes formatting fix, walkthru update

IanWhyte:


george-hoffer:
I'll fix the typo you added in ftps-and-edi-x12-purchase-order-walkthrough.adoc

IanWhyte:
You're making an assumption, George. In our meeting, I'll help you to
understand what it is so that we can deal with it in a more constructive
manner moving forward.

Ian

PR #832, opened 2016-09-29
First pass at API Policy clarification

ethanport:
This is a quick draft of some additional details about how policies are defined, deployed and executed.


krishahn:
Thanks much for the clarifications, Ethan.

PR #833, opened 2016-09-29
Elaborate section proxy deploy to PCF

nearnshaw:
Added specific details to the placeholder section I added about registering a proxy for PCF


krishahn:
Merged, Nico. Tks!

PR #835, opened 2016-09-29
Draft of more details on XA

ethanport:
Added details and clarified the XA transaction docs.  This should all be confirmed and checked with some experts. I just wrote some sample apps to verify the behavior I documented.


george-hoffer:
Thanks Ethan

PR #839, opened 2016-10-03
Shared Access Key level clarification

angelspano:
There's is an issue where trying to use Shared Access Keys at a lower level than the Namespace.
A Jira was created informing this SE-4493, until is fixed the doc update will help to avoid the issue.


george-hoffer:
Thank you Angel

PR #842, opened 2016-10-03

nearnshaw:
Added note about how the backend app needs to be on PCF for this to work.

There are several different use cases that can exist, and we need to clarify that we only support this one.


krishahn:
Sorry I missed your PR a few days ago.

PR #851, opened 2016-10-04
When you apply one of the policies, the others are greyed out. So, just one can be used at a time.

angelspano:


krishahn:
Merged, tks.

PR #858, opened 2016-10-07
Authorization Grant Types update in table

jmjdamen:
the grant types in the table were RAML 0.8, but in de example RAML 1.0 grants were pointed out
maybe change the table and add 0.8 AND 1.0 types as they are so different


krishahn:
Thanks!

PR #867, opened 2016-10-12
Proxy on PCF - last minute changes in tool

nearnshaw:
changed the code and the example file to match the latest changes in the tool


krishahn:
no need to use a PR, Nico. It's your content, so I don't care. Would be less trouble for you to just add what you want. :-)

PR #872, opened 2016-10-13
Add RHEL Mule init.d script

ethanport:


ethanport:
duplicated with another PR for 3.8

PR #878, opened 2016-10-17
Updated example and clarified inner workings

PabloCamino:
There was some confusion around if load balancer to worker communication needed to be encrypted or not, so clarified it and updated the example to also reflect it.


nearnshaw:
Thanks Pablo! I'm merging your changes!

PR #885, opened 2016-10-19
Add note that lookup doesn't support sub-flows

rhoegg:


nearnshaw:
thanks Rhoegg,  adding this to our doc!

PR #889, opened 2016-10-20
FTPS Connector v1.09 relnotes, add error codes & security, del endpoint files, add b2b trans sshot

IanWhyte:


george-hoffer:
Looks good. Thanks Ian.

IanWhyte:
You're welcome, George, thank you for processing.

PR #903, opened 2016-10-23
XML tag update for the domain example #7

jfielding:
The example domain project contains errors in the xml.  The closing xml tag should be `</domain:mule-domain>` instead of `</mule-domain>`.  The file is linked from https://github.com/mulesoft/mulesoft-docs/blob/master/mule-user-guide/v/3.8/shared-resources.adoc.


george-hoffer:
Thank you

PR #905, opened 2016-10-23
XML header and tag update for the domain example #9

jfielding:
Removing the outdated `<mule-domain>` tags and adding `<\domain:mule-domain> tag.  The file is linked from https://github.com/mulesoft/mulesoft-docs/blob/master/mule-user-guide/v/3.8/shared-resources.adoc.


george-hoffer:
Thank you

PR #925, opened 2016-10-27
Update clusters.adoc

adrianmartinmulesoft:
https://www.mulesoft.org/jira/browse/DOCS-1708

Added the multicast option


nearnshaw:
Thanks Adrian!

PR #928, opened 2016-10-27
Added STUDIO-8089

PabloCamino:
For some reason this bug-fix got lost, see https://www.mulesoft.org/jira/browse/SE-3856
Added it to the release notes, or customer won't allow us to close the case.


nearnshaw:
Thanks Pablo!

PR #929, opened 2016-10-27
Credentials for the Ops Center are email-based

pablodcar:
Also, adding how to reset the password of the the Ops Center


nearnshaw:
Thanks Pablo!
Merging these changes

Just a note, that I will correct now:  as we're white-labelling the gravity product, customers don't know what "gravity" is, we should always refer to it as a part of our solution

PR #934, opened 2016-11-01
Update engaging-users-of-your-api.adoc

PabloCamino:
As per https://www.mulesoft.org/jira/browse/AP-1670 this functionality is not supported


krishahn:
Tks

PR #935, opened 2016-11-01
Updated branding of Anypoint Platform and CloudHub

aaronlandgraf:
Removed articles "the" from Anypoint Platform mentions where appropriate.


george-hoffer:
Thank you Aaron

PR #941, opened 2016-11-04
chirag's requested change

krishahn:
Pls see request from Chirag in docs-feedback today. I changed my policy and provider doc to preface "external provider" with "Mule" as suggested by @evangelina. Here are similar changes I suggest for your doc, which was the specific one Chirag asked us to change.


fermujica:
This is perfect! Thanks for taking care of this request! 

Merging.

PR #944, opened 2016-11-04
Update backup-and-disaster-recovery.adoc

alejandrorascovan:


sharkberto:
Thanks Alejandro, implemented change locally and pushed, since GH showed not mergeable.

PR #945, opened 2016-11-04
on-prem instead of on-premise nor on-premises

ethanport:
In training we are using on-prem to represent any deployment into a customer managed infrastructure (on-premises or in another cloud environment). Never use on-premise because it's not a word. It would be on-premises, but again, customers do not need to deploy to Bare Metal.


fermujica:
The official term for documentation is 'Anypoint Platform On-Premises'.
We are not documenting the 'on-prem' expression.

PR #946, opened 2016-11-04
Clarify how to hide cloudhub.io in apps

ethanport:
Distinguish between VPC vanity domains using a dedicated load balancer vs. just adding your own DNS A records through your companies name servers.


fermujica:
Thanks Ethan! 
The term we are using in documentation is `On-Premises` instead of simply `On-Prem`.
Just to keep consistency, I'm merging your PR locally and apply some changes. After that, I'll push everything to master.

PR #952, opened 2016-11-07
Clarification on Quota Manager

PabloCamino:
I've just added a short paragraph explaining quota manager for throttling policies in a cluster and adding a recommendation not to use a very short time window. This comes after discussing with engineering for a couple support cases we had,


krishahn:
Thanks, much Pablo.

PR #953, opened 2016-11-07
SMTP Server must use a secure connection

pablodcar:
Only valid certificates (no self-signed) are supported for API Manager alerts throw SMTP.


cmordue:
LGTM

PR #956, opened 2016-11-09
Mule-Api-Version should be 1.0

ahnooie:
The Windows Gateway Service responds with "API version not specified or invalid" when Mule-Api-Version: "1".  Works with "1.0".


fermujica:
Thanks Benjamin! Your PR is now merged!

PR #957, opened 2016-11-09

federicoboerr:
PLEASE VALIDATE THE FORMAT AND THE NEW LINK BEFORE MERGING.

- Removed information about the Remote Assistance that is not relevant given that it will be disabled always (maybe we should remove from UI).
- Added new step to reset ops centre password


fermujica:
Thanks Fede. I just reviewed and merged this PR

PR #958, opened 2016-11-09
Update managing-schedules.adoc

angelspano:
Adding a note explaining the behaviour when more than one worker is available.


fermujica:
Thanks Angel! I just pushed your PR.

PR #962, opened 2016-11-12
moved SLO documentation from managing clients to managing users since…

cmordue:
… that is where it belongs


fermujica:
Thanks Chris!

PR #974, opened 2016-11-17
updated for proxy config so it is more clear what the format is after…

marcelbakker:
… the -P parameter


george-hoffer:
Thank you Marcel

PR #984, opened 2016-11-22
Minor clarification on Unmanaged

ethanport:


nearnshaw:
Thanks Ethan! Right you are!

PR #990, opened 2016-11-24
Deleted duplicated example block

adobni:
that was also incorrectly escaped.


fermujica:
Thanks Ale! 
I'm reviewing this PR because, for some reason, `creating-an-oauth-2.0a-web-service-provider.adoc` does not show up in our TOC. 
I'm checking internally to see if a newer version of this article is published instead.

PR #996, opened 2016-11-25
Update anypoint-filter-processor.adoc

adrianmartinmulesoft:
elaborate on IP filter due to feedback received in case https://na29.salesforce.com/5003400000smtam


adrianmartinmulesoft:
It seems that the | in the regular expression breaks the table. How can you have | escaped?

george-hoffer:
Entered manually with a unicode pipe symbol and edits.

PR #997, opened 2016-11-26
Add 1.6.0

ethanport:


nearnshaw:
Yes, we were also missing that link, thanks!

PR #998, opened 2016-11-26
Updated to the latest version 1.6.0 and links

ethanport:


nearnshaw:
Thanks Ethan! We were missing an update there, well spotted!

PR #999, opened 2016-11-28
note selinux must be disabled for ap onprem 1.5

sofuture:
cc @nearnshaw


nearnshaw:
Thanks Jeff! I just merged this

PR #1005, opened 2016-11-29
include commands to enable ap onprem mounts

sofuture:
@nearnshaw One more for you :)


nearnshaw:
Thanks Jeff!  Merging now!

PR #1006, opened 2016-11-30
Notes about onprem disk performance

sofuture:
cc @nearnshaw another one!


nearnshaw:
Thanks Jeff! Merging!

PR #1019, opened 2016-12-07
Cleaned up trivial spacing.

cgorshing:
Just a few changes where the spacing showed on the website and readability was hampered a bit.


fermujica:
Thanks for taking the time to fix this and improve our docs!

PR #1020, opened 2016-12-08

harirongali:
Fixing hostname field and description as it is confusing to users.


krishahn:
Thanks for the correction. @nearnshaw this pertains to a section you added to the API Manager docs. Please take a look. I'll merge.

PR #1027, opened 2016-12-13
Example of configuring and testing 2-way TLS

ethanport:


nearnshaw:
Thanks Ethan, nice addition!

I made some slight edits for it to render well from the asciidoc, but I didn't change any of the content, looks great!

PR #1042, opened 2016-12-20
by default http connector will be in error state

nathan-at-mulesoft:
The screenshots also need updating with these ones:
set payload
xml code


nathan-at-mulesoft:
(those images are to replace the ones in /images directory - but I don't have upload permissions).

george-hoffer:
Thank you Nathan

PR #1045, opened 2016-12-20
fix up incorrect maven link formatting

nathan-at-mulesoft:
and also used https link


nearnshaw:
Wow you've fixed quite a lot of these, thanks Nathan!

PR #1051, opened 2016-12-20
broken image / missing image

nathan-at-mulesoft:
Needs a new file uploaded (will include in pull request)
newmuleprojectgitignore


george-hoffer:
Copied image to doc, but left the old name. Closing without merging.

PR #1053, opened 2016-12-20
Improve awkward sentence

nathan-at-mulesoft:


nearnshaw:
Yup, that's certainly an improvement, thanks Nathan!

PR #1054, opened 2016-12-20
formatting to bold for UI menu selections

nathan-at-mulesoft:


nearnshaw:
Yes, that bold format makes more sense and matches our conventions used in other places, thanks!

PR #1060, opened 2016-12-20
fix broken link due to extra forward slash

nathan-at-mulesoft:


nearnshaw:
Yup, that's clearly a bug, thanks for spotting that!

PR #1061, opened 2016-12-20
broken url had localhost in it rather than relative link

nathan-at-mulesoft:


nearnshaw:
weird bug in that URL, thanks for spotting that!

PR #1073, opened 2016-12-22
set source to xml to (try) fix escaped characters

nathan-at-mulesoft:
My preview doesn't work properly, but I think this should work


nathan-at-mulesoft:
Hrm.. I'm not sure this will work now - as had a look elsewhere at the code and seems like it still messes it up?

nathan-at-mulesoft:
e.g. on this page: https://docs.mulesoft.com/mule-user-guide/v/3.8/splitter-flow-control-reference has similar problem at "Message Info Mapping" bit near the bottom. So I'm not sure on how to fix this if setting the source to type xml doesn't work properly.

george-hoffer:
Thanks for providing these fixes. I also fixed the Splitter Flow Control Reference per your recommendation. Thanks for pointing that out.

PR #1090, opened 2016-12-30
Advanced dataweave example

nuggetman:
I often have to demo this example, so am posting it up for all to use.


nearnshaw:
Thanks Alex, this is really helpful!  One of the most common feedbacks we get when we ask how we can improve our docs, is that we should include more examples for using DataWeave, so what you added here is extremely valuable to us :).

Just one note: When using DataWeave in Studio, the %input directive is not necessary to add, as the input is infered automatically from the context. I made a couple of tweaks to your pull request to remove this unncessary line of code and a couple of other minor fixes.

Thanks again!
Nico

PR #1094, opened 2017-01-02
Update installing-and-configuring-runtime-manager-agent.adoc

arielmira:
Clarification about which IPs/hostnames need to be whitelisted and how


nearnshaw:
Thanks Ari!  I'm adding the content you wrote!

Just one comment:  as we often put links to sections, we avoid putting characters like "/" in titles so that the links don't break.  I changed the title slightly so that we don't have "IPs/hosts". I'm also changing all the incoming links to this section on other pages so that they work with the new title.

PR #1099, opened 2017-01-04
fix sample url

IanWhyte:


fermujica:
Thanks for fixing this, Ian.

IanWhyte:
You're welcome - this page probably belongs under APM API, but the feature itself (business-process tracking) is still under development and it's not clear how it's going to work in the UI - so I think moving the page can wait.

PR #1100, opened 2017-01-05
Update authentication-in-http-requests.adoc

adrianmartinmulesoft:
Clarification on OAuth system property to override  redirectUrl


krishahn:
Thanks, merged.

PR #1106, opened 2017-01-09
Corrected simple connector typo

cgorshing:
Trivial changes - just found 3 instances where connector was misspelled.


george-hoffer:
Thank you Chad

PR #1108, opened 2017-01-09
Adding clarification on custom vs. default properties files

adrianpineyro:
Adding clarification on custom vs. default properties files


nearnshaw:
Thanks Adri!

PR #1123, opened 2017-01-17
Sw compat 3.8 upd

sharkberto:


sharkberto:
@dandiep in case you have any doubts or concers

PR #1126, opened 2017-01-18
Removing duplicate entry

adrianpineyro:
Removing duplicate entry and clarifying the exact version (SE-3802)


fermujica:
Thanks!

PR #1131, opened 2017-01-19
Fix capitalisation of a title.

mikecon94:


fermujica:
Thanks for catching this, Mike.

PR #1132, opened 2017-01-19
Update dataweave-flat-file-schemas.adoc

tonedef71:
Fixed minor typographical error (transposed characters).


george-hoffer:
Thank you

PR #1140, opened 2017-01-23
changes to some of the wording

nearnshaw:


sharkberto:
Thanks, looks great. I only have some minor add'l changes in return.

PR #1142, opened 2017-01-25
Tip for avoiding cases of non resolved properties in tests

WilliamPinho:


fermujica:
Thanks Will, good tip.
I took the liberty of adding your tip to the one above so all system properties tips could be found in one place. 
Let me know what you think.

WilliamPinho:
Thanks, Fer! Yes, that way is better.

PR #1150, opened 2017-01-27

PaulAndersonMule:
removed "already" from "should have Anypoint Platform already running"


nearnshaw:
Ok, thanks Paul! Adding your changes!

PR #1152, opened 2017-01-28
Changed to valid JSON on the examples

wpoch:


nearnshaw:
Thanks Walter!  merging now

PR #1159, opened 2017-02-02
Update twilio-connector.adoc

gramone:
The proposed change allows creating a global element for the Twilio connector that will pass schema validation at runtime. The documented step details how to complement the XML generated by the global element creation wizard.


george-hoffer:
Thank you Gilles

PR #1161, opened 2017-02-03
Update prerequisites-platform-on-premises.adoc

PaulAndersonMule:
extra clarity on device-types


nearnshaw:
Thanks Paul!  I'm adding your changes.
Besides the improved clarity, in our docs we never use the word "we", which you also keenly corrected :)

PR #1163, opened 2017-02-03
Update setting-smtp-manually.adoc

facunlk:
the argument should be -d


fermujica:
`> base64 --help` 
Returns:
`-D, --decode   decodes input`

`> echo "MWYyZDFlMmU2N2Rm" | base64 -d` returns
`base64: invalid option -- d`

Is it working with the -d option on your end?

facunlk:
Hello Fer,

Sorry for the delayed answer. That option is like that in Mac but in Red Hat/CentOs where on-prem is actually installed the option is with lowercase.

`[anypoint@anypoint1 ~]$ cat /etc/centos-release
CentOS Linux release 7.3.1611 (Core)
[anypoint@anypoint1 ~]$ base64 --help
Usage: base64 [OPTION]... [FILE]
Base64 encode or decode FILE, or standard input, to standard output.

Mandatory arguments to long options are mandatory for short options too.
  -d, --decode          decode data
  -i, --ignore-garbage  when decoding, ignore non-alphabet characters
  -w, --wrap=COLS       wrap encoded lines after COLS character (default 76).
                          Use 0 to disable line wrapping

      --help     display this help and exit
      --version  output version information and exit

With no FILE, or when FILE is -, read standard input.

The data are encoded as described for the base64 alphabet in RFC 3548.
When decoding, the input may contain newlines in addition to the bytes of
the formal base64 alphabet.  Use --ignore-garbage to attempt to recover
from any other non-alphabet bytes in the encoded stream.

GNU coreutils online help: <http://www.gnu.org/software/coreutils/>
Report base64 translation bugs to <http://translationproject.org/team/>
For complete documentation, run: info coreutils 'base64 invocation'`

fermujica:
Thanks! I didn't know this.
This PR currently conflicts with the existing private cloud docs because the section has been restructured.
I see that the current docs are using the `-d` option. 

Thanks for flagging this initially.

PR #1164, opened 2017-02-03
Update managing-api-clients.adoc

aj90:
[NOTE]
--
External IdP integration for Client management is only available to our Platinum subscription users only. (Enterprise edition is a deprecated term )


george-hoffer:
Thank you Ashley

PR #1165, opened 2017-02-05
Customer Success request

krishahn:
@george-hoffer letting you know about this change to the Exchange docs (you handle these, I think). Please merge asap. As I think you know, Customer Success requested that we make this info very visible quite a while ago, so this fix is coming in too late. Sorry to have waited so long.


george-hoffer:
Thank you Kris

PR #1169, opened 2017-02-08
Added clarifying statements to External Identity that these settings applies to all business groups.

cmordue:
We've had a couple questions related to this recently.


fermujica:
Thanks Chris!

PR #1183, opened 2017-02-21
Clarify the Next Steps

ethanport:
Just added a little more context so they are easier to read and understand.


george-hoffer:
Thanks Ethan!

PR #1187, opened 2017-02-23
Update creating-a-policy-walkthrough.adoc

manibh:
user need to have Organization Administrators role in order to be able to create custom policy and there is no API administrator permissions at all.


krishahn:
Thanks.

PR #1189, opened 2017-02-23
Update microsoft-dynamics-crm-connector.adoc

NathanNam:
Add a new version of the supported CRM Online


george-hoffer:
Thanks Nathan

PR #1191, opened 2017-02-24
Added more runtime versions that are now supported

vnath:
Added more runtime versions that are now supported


george-hoffer:
Thanks Vivin

PR #1194, opened 2017-02-24
Runtime Manager is wrongly named Cloudhub in Mule docs.

FGambarg:


fermujica:
Indeed! Thanks for catching this Facundo.

PR #1195, opened 2017-02-24
dep target clarification

sharkberto:


nearnshaw:
Awesome, I like what you did there Alex! It reads out a lot clearer now :)

PR #1197, opened 2017-02-28
Last section / client secret is still generate by Anypoint not PF

jbugnet:
The client id/secret is generated by anypoint code, then the client is created in PF using this POST call: https://documentation.pingidentity.com/display/PF72/OAuth+Client+Management+Service#OAuthClientManagementService-1082697
Saying you get the client id and secret from PF is misleading


krishahn:
Thanks.

PR #1206, opened 2017-03-04
Update design-an-api.adoc

vinnunaidu:
screen shot 2017-03-04 at 15 53 30
Removed the double quotes for 'example' property in queryParameters, as the type is an Integer. API Manager complains about this error.


george-hoffer:
Thank you for your contribution to MuleSoft documentation. For future fixes, please complete the contributor's agreement at https://github.com/mulesoft/mulesoft-docs/blob/master/README.adoc to register as a contributor so that we can accept your pull requests. Thanks much for helping improve MuleSoft docs!

PR #1209, opened 2017-03-06
Update custom-log-appender.adoc

adrianmartinmulesoft:
Clarify that the log files are not downloadable from the UI when custom-log-appender is enabled

Also, removed the "This will be addressed in a future release." because we don't have a date for ETA


nearnshaw:
Thanks Adrian!

PR #1219, opened 2017-03-08
adding binary and addressing boolean

adrianpineyro:


nearnshaw:
Thanks Adri!  Well noticed!

PR #1221, opened 2017-03-08
Update database-connector-examples.adoc

PaulAndersonMule:
having a huge number of rows of example data is not instructive - it obfuscates the true intent which is to explain information


george-hoffer:
Thanks Paul

PR #1233, opened 2017-03-11
small clarifications on the split function

ethanport:


nearnshaw:
Thanks Ethan! Merging now

PR #1234, opened 2017-03-13
Update dataweave-memory-management.adoc

gustavo-caracoglia:
- For SE-5517, added the DataWeave limitations for:
    - the size for the whole payload
    - the size for single fields


nearnshaw:
Thanks Gus!  Merging now!

PR #1236, opened 2017-03-13
Removed unused pattern.

wpoch:
Now every field is published using a JSON object.


nearnshaw:
Thanks Walter!

I just renamed this doc to "sending data from arm to external ANALYTICS software" (instead of monitoring). So I can't merge your change. Sorry but I'll have to close the pull request.

I just copied your changes manually to the new file and pushed separately.

PR #1237, opened 2017-03-13
Fixed internal ID for hazelcast reference

ethanport:
Binding to Hazelcast


nearnshaw:
Thanks Ethan!  The reference link and the link text were mixed up there, good catch! :)

PR #1245, opened 2017-03-17
index param optional in map operator

ethanport:


nearnshaw:
Thanks Ethan!

Just a side-note: Mariano confirmed to me that in Mule 4, the map operator only works with arrays. For mapping objects, you use mapObject. Given that, it might be best to not put too much enfasis on the use of 'map' on objects to avoid future incompatibility problems.

PR #1253, opened 2017-03-22
Deploy config on maven plugin modified

jorgegarciamule:
The example was changed because it was not an example and syntax was confusing.


nearnshaw:
Thanks Jorge!
Adding your changes

PR #1261, opened 2017-03-27
Remove duplicated bodyStartLineNumber definition

adobni:


nearnshaw:
Thanks Ale!  Yes, that's duplicated, good find!

PR #1264, opened 2017-03-28
Update twilio-connector.adoc

NathanNam:
Added "Account SID" to username and "Auth Token" to password.


george-hoffer:
Thank you

PR #1279, opened 2017-04-03
Proxy must be configured in both, wrapper.conf and mule-agent.yml

damiancalabresi:
The connection to Runtime Manager Console must be configured in the wrapper.conf file.
But since Runtime Manager Agent 1.5.1 the connection to the Auth Proxy endpoint, that the Agent uses to send the Insights and the server monitoring retrieves the proxy information from the mule-agent.yml file.

The wrapper.conf configuration doesn't replace the one in the mule-agent.yml.


nearnshaw:
Thanks Damian! I'm merging your changes now!

PR #1280, opened 2017-04-03
Update rosettanet-connector.adoc

dsosnoski:
Add details and clarify how location ID values are now used.


george-hoffer:
Made changes manually so as not to overwrite a new section. Closing. See https://docs.mulesoft.com/mule-user-guide/v/3.8/rosettanet-connector for more information after 3:30pm 3/Apr.

PR #1281, opened 2017-04-04
Implement Mule Message and Mule Event structures into beta docs.

fermujica:
This PR adds:

Mule Message Structure section to the MUG _toc.
Mule Event Structure graphic to mule-concepts.adoc

It also re-words the following from mule-concepts.adoc:

> A Mule event is generated by the **Event source** of a flow, which is configured to be triggered by a _resource_ that might be external to the Mule Runtime, such as an HTTP request, or a change to a database or file.

For:

> A Mule event is generated when an _external event_ (such as an HTTP request, or a change to a database or file) reaches the **Event source** of a flow. This _external event_ can be triggered by a _resource_ that might be external to the Mule Application.

It also fixes some broken anchored links.

See the commit message in each commit for more information.

Let me know if this looks good for you.


fermujica:
Thanks for your review, @sharkberto. 
I noticed the extra `_See Also_` in the beta docs. I have no problem in separating them in different **About** topics if it improves our searchability. 

Regarding your inquiries about why I'm focussing on the Mule Event rather than the message alone: The message in Mule 4 is different that the Mule message in Mule 3.x. All information relevant to the flow execution (for example, the correlationId) is now part of the Mule event , not the message (as it used to be in 3.x).
The message in Mule 4 is simpler and only represents the value being processed along with its data type and related meta-data. 

Think of the Mule message as an operation's result, and the Mule Event as the target for transformation.
If a user wants to keep persistency of an operation result (a new message), each operation allows the user to set a target for this result. This target is a variable, external to the mule message.. but part of the Mule event. 

I want to make clear that the user now has complete control over the context of information around a message.. but not over the message itself. 
Now, how to access this data using dataweave is something I don't know for sure yet, but it would definetely help to show some examples here. 

The logic behind mentioning dataweave as a core concept for the runtime is because it is the runtime's expression language and a core component of it. I think we should at least mention it, giving a quick glimpse about why it's useful and then link to the specific dataweave documentation.  But we can discuss this topic further though.

PR #1291, opened 2017-04-06
Added the missing example for /orders post method

sureshchoudhary:


sureshchoudhary:
My bad. Added the wrong example.

PR #1292, opened 2017-04-06
Added the missing example for /orders post method

sureshchoudhary:
This example is missing in the earlier example and suddenly pops up later in the full RAML, confusing the user.


george-hoffer:
Thank you Suresh!

PR #1296, opened 2017-04-07
Removed the `per minute`

wpoch:
The lapse of the metric depends on the period of time you are selecting.


nearnshaw:
Thanks Walter!

PR #1301, opened 2017-04-11
Update anypoint-platform-cli.adoc

adrianmartinmulesoft:
put warning that CLI doesn't support External Identity Users. Base on support case and https://mulesoft.slack.com/archives/C2PLN0FRB/p1490117463912861


george-hoffer:
Thank you Adrian!

PR #1302, opened 2017-04-11
update basic tutorial app address

GustavoARod:
As a matter of avoiding confusion and making easier to follow, the basic-studio-tutorial should indicate to type the 127.0.0.1:80801 address in the browser instead of 0.0.0.0:8081


GustavoARod:
Hi Fer,

It's my pleasure to make any contribution to such a nice project like Mule.

My point is that I consider there is a misunderstanding between the listener configuration and the actual IP address that Mule runtime is running by default.  I think that the host in the listener configuration refers to the IP addresses that an application is going to be able to listen (from a security standpoint. On the other hand, the IP address I'm talking about is the one that will be used by Mule runtime to run my application, which is localhost by default.

So, anyone that is just simply following the step by step carefully, as it was my case, won't be able to see anything besides 'This site can’t be reached' on the browser as soon as enter has been pressed, only because nothing is running on that IP address 0.0.0.0. Nevertheless, the user will be able to see his/her application working if it types localhost:8081

I hope this could help make my point a little bit clearer.

Thanks.

george-hoffer:
Hi Gustavo, The Basic Studio Tutorial works correctly with the 0.0.0.0:8081 address in the browser. Please try the tutorial and see. We test our docs and I can assure you that this tutorial works correctly. When the HTTP connector is configured, 0.0.0.0 is the default. In addition, the 0.0.0.0 address is required when exporting an app to Runtime Manager (aka CloudHub). 

We do appreciate your contribution and that you want to improve MuleSoft docs. 
We would appreciate if you would close this PR request. Or we can close it for you in a few days.

For future contributions, please complete the contributor's agreement as described at https://github.com/mulesoft/mulesoft-docs/blob/master/README.adoc to register as a contributor so that we can accept a pull request. Thank you.

GustavoARod:
Thanks, @george-hoffer! I just closed the PR

aiannucci:
@fermujica @george-hoffer, can we review this documentation?
Even when our recommendation to any user is to configure an HTTP connector binding to 0.0.0.0, in order to bind to all interfaces, you can't make a request to 0.0.0.0
0.0.0.0 is not an address, is a keyword designed to bind to all local interfaces. But when you make a request, you need to do it to an specific address, you can't do it to 0.0.0.0
That being said, I know no browser where you can do a request to 0.0.0.0 and it works.
Have you tested this in any environment?

george-hoffer:
@aiannucci Yes we test our docs and I assure you that you can trigger an app in a browser with 0.0.0.0:8081. One of the larger reasons we encourage the use of address 0.0.0.0 is that this address is required when deploying an app in Runtime Manager. Our doc policy is to stick with 0.0.0.0 in our examples so that apps can easily be ported between Studio and Runtime Manager. As you are an employee, please discuss this issue in Slack vs this venue.

PR #1305, opened 2017-04-12
Update cloudhub-fabric.adoc

jmjdamen:
Added 'CloudHub' for column


nearnshaw:
Thanks Joost!
I reworded things a little, because we want to aviod speaking of "CloudHub applications". We've had feedback about that being confusing to new users (they think they're a kind of application different from Mule applications), so we've been rewording that all throughout our docs.
Merged!

PR #1325, opened 2017-04-21

rahuldureja:
Updated the Using Remove on Array example from ["a","b","c"] - 1 to ["a","b","c"] - "b"


nearnshaw:
Thanks Rahul
I'll make the same changes on other versions now

PR #1329, opened 2017-04-23
Removed incorrect DW expressions

ethanport:
Show mix of empty and non-empty input to see how the directive applies only to empty values.


nearnshaw:
Thanks Ethan, That looks a lot better!

I also don't understand what that last line `emoveAttribute(payload, "password")` was, especially as there was no "password" field in the original example... weird. Good thing you got rid of it!

PR #1333, opened 2017-04-26
fix a misplaced apostrophe

nathan-at-mulesoft:


nearnshaw:
Thanks Nathan! Adding your fix

PR #1334, opened 2017-04-26
remove the talk about MMC which is now deprecated

nathan-at-mulesoft:


nearnshaw:
Hi Nathan,
That link you added for downloading MMC http://www.mulesoft.com/managing-mule-esb-enterprise
is redirecting to https://www.mulesoft.com/platform/integration-manager, which is about Runtime Manager.
Perhaps this is a deliberate change from marketing, since we want to encourage people to use Runtime Manager rather than MMC. In any case, if the link takes you to a different product, that's more confusing than helpful.  Any ideas what the new link for that would be? I'm not finding anything on MMC in that marketing page.

PR #1336, opened 2017-04-26
Fix up out of date link with always current link

nathan-at-mulesoft:


nearnshaw:
Thanks Nathan!  It makes a whole lot more sense to link to the parent page as you just did, one thing less to worry about keeping up to date :)

PR #1363, opened 2017-05-03
Clarifying what are the options for the log path

germansolis:
We need to clarify this as It creates some confusion on the customer, who could try, for instance, to use paths relative to the app's classpath


nearnshaw:
Thanks German!  Adding your changes!

PR #1372, opened 2017-05-05
Add a clear warning about using maxBufferSize

adobni:
This change should be pushed to previous Mule versions of this page.


fermujica:
Thanks Ale. I replicated this note to older versions as well.

PR #1377, opened 2017-05-06
repair more broken links

mojavelinux:


mojavelinux:
With these fixes along with updates I made to the spreadsheet / migration sheet, we should be down to just a handful of broken references.

nearnshaw:
Thanks Dan! Merging your fixes

PR #1380, opened 2017-05-08
Update deploying-to-cloudhub.adoc

manibh:
adding following one as it was not clear how much storage is attached to /tmp for 1 vCore workers.
* 1 vCores workers have an additional 4 GB of SSD storage mounted on /tmp


george-hoffer:
Thank you Mani

PR #1381, opened 2017-05-08
Update external-oauth-2.0-token-validation-policy.adoc

marcelbakker:
Please add the fragment with the proxy config if behind a reverse proxy which explains the addition of the anypoint.platform.external_authentication_provider_enable_proxy_settings to the configuration of Mule RT


krishahn:
tks

PR #1382, opened 2017-05-08
remove hearing a response in 48 hrs

cmordue:
Update docs to remove about hearing from mulesoft within 48 hrs of form submission since this is now self-serve and mulesoft support is no longer involved in entering/validating their configuration.


george-hoffer:
Thank you Chris

PR #1392, opened 2017-05-12
fix ref to VPC size section

mojavelinux:


fermujica:
Thanks for fixing this. This link was pointing to a different page that's not built yet.

PR #1393, opened 2017-05-12
fix link to contacts.raml attachment

mojavelinux:


mojavelinux:
Does this look correct @fermujica?

fermujica:
Yes. It's correct.
Thanks !

PR #1397, opened 2017-05-15
Update creating-an-odata-api-with-apikit.adoc

gramone:
This attribute must exist and be set to true for the flow to work properly at runtime.


krishahn:
tks

PR #1402, opened 2017-05-17
Update logging-in-mule.adoc

germansolis:
Correction to the config's path resolution


nearnshaw:
Thanks German!  Adding this change in!

PR #1410, opened 2017-05-20
fix AsciiDoc syntax errors that pertain to image macros

mojavelinux:
This patch contains AsciiDoc syntax corrections that cannot be safely fixed using the migration script. 

The purpose of the patch is to either convert an inline image into a block image, or put it in a place that allows it to be converted automatically by the migration script.


sharkberto:
@sleslie23 @dukesphere  I would like to merge this as soon as possible. Note there are some changes to files in /runtime-manager. Could you confirm I can merge or perform the merge?

mojavelinux:
To provide some context, isolated inline images are being converted to block images in the new site. However, there are cases that we cannot catch with the script. Those cases I've changed manually.

When changing to block images, the list continuations also have to be correct. In certain cases, those changes have to be applied manually, which is also part of this fix.

In a nutshell, this fix aims to correct the AsciiDoc to align with the behavior that was intended.

PR #1406, opened 2017-05-18
removed url

jmjdamen:
this should not be here i guess?

PR #1411, opened 2017-05-20
fix incomplete link macros

mojavelinux:
Prompted by #1406, I found many more cases of incomplete link macros.

- superfluous link: prefix
- missing square brackets
- misplaced square bracket

Supersedes #1406.


mojavelinux:
Please note that the `link:` prefix is not required for a qualified URL (a URL that begins with a scheme). The migration script will drop the prefix where not needed, but it misses the cases when the macro is not fully formed.

PR #1416, opened 2017-05-24
Updating /tmp storage requirements

jfielding:
Based on a recent POC of 1.5.2, the /tmp directory requirements should be expanded from 7GB to 10GB.


sleslie23:
Thanks @jfielding!

PR #1422, opened 2017-05-26
Update index.adoc

manibh:
updated the mule runtime version and added 3.8.3 as per Munit 1.3.2 release note:
https://docs.mulesoft.com/release-notes/munit-1.3.2-release-notes

But needs to clarify if it also supports 3.8.4 as well for both release note document update and this page.


george-hoffer:
Thank you Mani

PR #1426, opened 2017-05-26
Add link to full match operator description

ethanport:
This is in the Language Introduction (which is confusing, since this is a core operator. We at least need to link back to the information).


nearnshaw:
Thanks Ethan!
Adding this to the beta version of 4.0 docs too!

PR #1431, opened 2017-05-30
Update runtime-manager-agent-1.7.0-release-notes.adoc

adrianpineyro:
This has not been tested, as confirmed by the Engineering Manager so it needs to be removed.


sleslie23:
Thanks @adrianpineyro !!!

PR #1440, opened 2017-06-01
Updating button label for adding items to Exchange

jfielding:
The label for the button in Anypoint Exchange is "Add Item" instead of "Submit Item"


george-hoffer:
Thank you Jay!

PR #1454, opened 2017-06-02
change for batchSize detail

adrianmartinmulesoft:
The maximum number is 10, if a higher value is used 10 is forced.


george-hoffer:
Looks good. Thanks for adding this.

PR #1464, opened 2017-06-02
fix internal links flagged by migration script

mojavelinux:
When a page wasn't available in the same version, I referred to the closest version that contained the referenced page.


sleslie23:
Thanks @mojavelinux !!!

mojavelinux:
:+1:

PR #1470, opened 2017-06-06
fix broken images flagged by the migration script

mojavelinux:


mojavelinux:
Most of these are removing a non-existent thumbnail for a video and changing the video from a link to an embed.

sleslie23:
Thanks @mojavelinux !!

mojavelinux:
:+1:

PR #1471, opened 2017-06-06
fix AsciiDoc syntax errors flagged but not handled by migration script

mojavelinux:


anypoint-private-cloud/v/1.6/config-alerts.adoc
```diff
@@ -7,11 +7,11 @@ Anypoint Platform Private Cloud Edition sends alert emails when CPU and memory u
 
 To setup alerts:
 
-1. Configure your SMTP email server as described in link:/access-management/private-cloud-edition-features#smtp[SMTP].
```
mojavelinux: @sleslie23 you seem to be creating ordered lists using Markdown syntax. Please see http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#lists. In short, you don't need a number or letter before the dot. The number of dots equals the level of nesting.

sleslie23:
@mojavelinux I "seem" to be creating them because in the past they have worked and I find that the "." syntax for lists is impossible to read. If that is required in the new system then that is what I will use.

Please don't submit PRs. I'll change them myself. Thanks.

sleslie23:
@mojavelinux Hey Dan. I fixed the list formatting in all of the files under the private cloud directory. I will also fix these in the runtime-manager reorg we are currently working on.

mojavelinux:
That's how ordered list items have always been defined in AsciiDoc. You can put the numbers, but then you have to maintain the sequence (as the sequence is validated). Markdown ignores the numbers, which is why I was suggesting that might have been the scheme you were following.

I added that change to this pull request because we are trying to get all warnings eliminated from the migrated files...and that was the only place it was warning. I'm glad to hear that you fixed the others. Thanks!

mojavelinux:
There are many other syntax fixes that need to be made that were part of this branch. If I rebase, can you reopen and have someone review and merge it?

mojavelinux:
I went ahead and just created a new PR.

PR #1476, opened 2017-06-06
Add explanation for empty logger

adobni:
Change the previous example to actually print the payload instead.


nearnshaw:
Thanks Ale!
This is useful information!  I reworded it a bit so it's clearer, and gave it titles that I think will make it easier to find.

PR #1477, opened 2017-06-06
fix AsciiDoc syntax errors flagged but not handled by migration script

mojavelinux:


mojavelinux:
@sharkberto These are syntax fixes that cannot currently be fixed by the migration script (in other words, they are the exceptions).

sleslie23:
@mojavelinux Hey Dan! Alex is on vacation, so @mention me if you need anything merged quickly.

mojavelinux:
Got it. Thanks @sleslie23!

PR #1478, opened 2017-06-07
Fix capitalisation on a title in DataWeave Operators.

mikecon94:


fermujica:
Thanks for fixing this, Mike.

PR #1485, opened 2017-06-08
Create mq-connector-release-notes.adoc

adrianmartinmulesoft:
Version 1.1.1 introduced a bug that affects that overwhelm the platform causing a DoS. SE-5286.


george-hoffer:
Thanks Adrian!

PR #1491, opened 2017-06-09
Update pointcuts-reference.adoc

PabloCamino:
Corrected resource level XML as it had an extra line


george-hoffer:
Thank you Pablo

PR #1504, opened 2017-06-13
describe firewall now created rules on port range

sahilmadan85:
While creating a test vpc I noticed that I ended up with 2 rules instead of 4 with port ranges 8091 to 8092 and 8081 to 8082.


fermujica:
Yes, I fixed this in another branch that I still need to merge and forgot to update it in the Master.
Thanks for doing so!

PR #1505, opened 2017-06-13
Clarified the profile and server setting changes

adampridmore:
The server and profile instructions aren't as clear as the other sections that the settings should be in a servers and profiles element.


george-hoffer:
Thank you Adam!

PR #1506, opened 2017-06-13
Changed download link

cmartinezledo:
We are actually pointing to a beta version instead of the released version


fermujica:
Good point. Thanks for noticing this. I might have copied a beta link without noticing it.

PR #1513, opened 2017-06-18
Added disconnected RunTimes scenario

anweshrijal:
Added information about what happens if communication between RunTimes and Management pane is disrupted.


george-hoffer:
Thank you Anwesh!

PR #1514, opened 2017-06-18
17q2 tls dangling

nuggetman:
I do not know why this is showing as a diff "api-manager/v/latest/browsing-and-accessing-apis.adoc" when I did not touch the file and it was pulled from the Mule master. In any event there is not change there.

The change is to the TOC for the Mule User Guide to add a dangling page.


george-hoffer:
Thank you Alex!

PR #1515, opened 2017-06-19
grammar fix

jmjdamen:
grammar fix because the previous one seemed wrong, use this one or change to another


george-hoffer:
Thank you Joost!

PR #1517, opened 2017-06-20
Update upgrade.adoc

adrianpineyro:
This file is not available in the Support Center. It is provided by a service professional who leads the installation.


sleslie23:
Thanks @adrianpineyro !!!

PR #1520, opened 2017-06-21
Update flow-processing-strategies.adoc

derekXlin:
put extra information in explaining what the "supported components" means, it is misleading as customer would think all the listed components are supporting "None-Blocking" operations


nearnshaw:
Thanks Derek, that clarification makes sense to include there. I'll merge your changes!

PR #1521, opened 2017-06-21
17q2 api auto disco

nuggetman:
Currently there is nothing documented regarding manually deploying to CloudHub using auto-discovery, these documented steps are necessary to make it work.


krishahn:
pls see suggestion to add a pointer in the auto-discovery section to Deploying to CH where we propose adding a new section about setting the properties to include client id and secret.

PR #1526, opened 2017-06-21
Update organization.adoc

aj90:
Trial accounts have access to Business groups. This is incorrect: 
[WARNING]
Business Groups is a feature that isn't available by default in a new Anypoint Platform account. To activate this functionality in your organization, contact your MuleSoft representative.


fermujica:
This seems to be an old note. 
Thanks for updating it,  Ashley.

PR #1528, opened 2017-06-22
Update microsoft-dynamics-365-connector.adoc

adrianmartinmulesoft:
Changed Oauth2.0 limitation. It generated confusion with the customers


george-hoffer:
Thanks Adrian!

PR #1540, opened 2017-06-27
Fix link for Studio "hardware and software requirements"

the-jude-abides:
Link should point to Studio requirements, not Mule runtime requirements. https://docs.mulesoft.com/anypoint-studio/v/6/hardware-and-software-requirements 

As it is, the Studio requirements page appears to be an orphan, with the current link instead pointing to the Runtime requirements.


nearnshaw:
Thanks Jude, this change makes a lot of sense!

In the past, that doc in the runtime section used to contain info about studio hardware and software reqs as well. As our product offering expanded, we started partitioning hardware and software requirements into different sections per products. It seems like we forgot to point that link to the new place where we have that content.

PR #1542, opened 2017-06-27
Update internal-handler-buffering.adoc

farolfo:


nearnshaw:
Thanks Franco!  
Merging with some changes to the wording. I also changed the wording in the table of the maximum buffer size parameter so that it explicitly says that it's measured in number of events.

PR #1557, opened 2017-06-30
Changed to YAML the examples

wpoch:


george-hoffer:
Hi Walter -- Why do you want to make this change?  We typically cast code examples with [source] or sometimes just ---- above and below the code. We haven't used 3 tick marks in other docs.

nearnshaw:
Fixed the syntax so that it matches what we always use, merging now. Thanks Walter!

PR #1577, opened 2017-07-11
Update install-create-lb.adoc

federicoboerr:
The health endpoint for the load balancer is HTTP not HTTPS.


sleslie23:
Thanks @federicoboerr !!!

PR #1580, opened 2017-07-12
Added FAQ

adrianmartinmulesoft:
a question received by a customer: Why I see the message decripted in my encrypted queues?

@george Please check the `desencrytion` word. Not sure what's the correct english form


george-hoffer:
Thank you Adrian!

PR #1586, opened 2017-07-13
Added Note about retention period

the-jude-abides:
Per @walter.poch , the retention period for flow metrics is two weeks. I wasn't able to find this information anywhere other than a response on Slack, so it should be documented here. Added a Note to that effect.


sleslie23:
Thanks @the-jude-abides !!!

PR #1587, opened 2017-07-14
Fixed transpose of binary and boolean types

adampridmore:


nearnshaw:
Thanks for your contribution Adam!  It seems that someone else has spotted this same mistake and fixed it already. These links now work well in the page.

I'll close this pull request, but thanks anyway for taking the time to send this!

PR #1597, opened 2017-07-19
updating Jersey client link

the-jude-abides:
Old Jersey client link is dead, long live the new Jersey client link.

Alternatively, instead of linking to the main page, we could link to the (more likely to change) download page: https://jersey.github.io/download.html


fermujica:
Thanks for updating the link!

PR #1605, opened 2017-07-21
Update http-about-http-request.adoc

afelisatti:
This doc was mixing up the request and listener functionality. I've edit it to consider only the listener (mainly because of the featured image). We'll need a similar doc for the request component.


afelisatti:
Also, the image has an error regarding the access to the Host header which it shows repeated as the URI expression.

krishahn:
Thanks Ana, there was a merge conflict so I put your changes in from the diffs. https://beta-anypt.docs-stgx.mulesoft.com/connectors/http-about-http-request Also fixed the image (that change will appear on the beta site in a few minutes) to read attribute.request.host.

PR #1608, opened 2017-07-21
Changing "Eclipse" for "Anypoint Studio"

fermelone:


fermujica:
Thanks! if it says Eclipse it must be really old. I'm replicating this change to 3.7, and previous versions as well.

PR #1622, opened 2017-07-24
Update starting-and-stopping-mule-esb.adoc

jmjdamen:
status is a Unix only command


nearnshaw:
Thanks Joost! Adding now

PR #1627, opened 2017-07-24
Remove masterfile redeployment mention

jmjdamen:
All XML configuration files mentioned in the deployment descriptor are monitored for changes and will trigger a redeployment when the setting is set to true.
Not sure when this was introduced but I tested this and this is the current behaviour for 3.8


george-hoffer:
Thank you for this fix!

PR #1642, opened 2017-07-27
Fixed the POM in the runtime agent extension tutorial

ashosborne1:
Added a POM that will allow someone to extend the runtime agent correctly. Also added a call out that java 1.7 must be used to compile the extension or it will not be loaded


nearnshaw:
Thanks Ashley!  I'm adding your contributions to our doc!

PR #1645, opened 2017-07-28
Minor changes to landing page

amitsaxena24:
First para and then edited language for Custom asset type.


george-hoffer:
Thanks Amit

PR #1654, opened 2017-07-31
Added note about the URL change

amitsaxena24:
added migration topic link as well.


nearnshaw:
Thanks Amit.
I made some small edits to the text and placed it in its own section about migration so that it stands out more.
I'm merging it into master now

PR #1661, opened 2017-08-01
Update MUnit Assertion Docs

diegostrubolini:
We should also clarify that the is parameter receives only a matcher


fermujica:
Thanks Diego!

PR #1667, opened 2017-08-01
Update arrange-cards-flow-design-center.adoc

Mrk-Nguyen:
There are no Scopes component in flow designer at the moment. Does the document mean scopes within the Try and Validation components?

### DIFF
--- a/design-center/v/1.0/arrange-cards-flow-design-center.adoc
+++ b/design-center/v/1.0/arrange-cards-flow-design-center.adoc
@@ -4,7 +4,7 @@
 
 Flows are composed of components that are chained together, where each component sends its output to the next component. Each of these components is represented in Design Center as a card.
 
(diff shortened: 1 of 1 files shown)


nearnshaw:
Thanks Mark, I like all of these improvements you're making!

About scopes, this is another topic we're in discussion with the UX team about. Try, For Each, etc have traditionally always been known as "scopes". In the Studio UI, they are found under the "scope" category. But it is true that there is no mention of the concept of "scope" in the Design Center UI. 

It is correct to refer to them as "components", as everything is considered a component in Mule 4. On the other hand, a "scope" could be considered a subgroup of "components", and the distinction could help clarify. Although I also get how the distinction could create confusion... it's something to be debated further.

nearnshaw:
I enumerated a couple of scopes in the explanation to make it a bit clearer.

Mrk-Nguyen:
Thanks for the explanation on scope discussion you're having with the UX team. In the mean time, I'll refer Try and For Each as components in my training manual for flow designer.

PR #1669, opened 2017-08-01
Update to-set-up-connector-configurations.adoc

Mrk-Nguyen:


nearnshaw:
Thanks Mark, I like these changes. Your example is easier to grasp too :)

PR #1670, opened 2017-08-01
Update run-app-design-env-design-center.adoc

Mrk-Nguyen:


nearnshaw:
Thanks Mark, your edits are well in-line with the editorial style we want for these topics :)

PR #1673, opened 2017-08-02
say "flow elements", not "building blocks"

ethanport:
Seems like we are moving away from "building block" to just say "flow elements" to include message processors and message sources together.


fermujica:
Thanks Ethan! 
We are focusing our efforts in unifying the terminology we use among products. 
We are definitely getting rid of the "Building Block" term. 
That said, I don't think we should introduce a new term such as "Flow element". 
I had a few ideas:
Since "component" is a defined term in our glossary, we can call them that, or even I would go for "flow component". 
Even simpler, I'd define these as processors (in Mule 3.x we have message processors, and in Mule 4.x we have event processors).

I'm closing this PR and continue working in #1693

PR #1675, opened 2017-08-02
fix asterisk that was getting formatted as bullet

nathan-at-mulesoft:


nearnshaw:
Well spotted, thanks!  Merging now

PR #1685, opened 2017-08-02
fix up some inconsistencies in the terms/language

nathan-at-mulesoft:
foreach variants purged in favour of "For Each Scope"


nearnshaw:
Thanks Nathan!  These changes make a lot of sense

PR #1686, opened 2017-08-02
Consistency for the sub elements

nathan-at-mulesoft:
Added in "one of:" for several elements, in line with how the others were defined


krishahn:
Nathan, thanks for attempting to fix this doc. The technical references (one for each connector) is generated from Javadoc comments written by the developers. The docs team isn't editing them--no point to doing that because as soon as there's an update, the edits (yours in this case) will be wiped out. The idea behind using the generated docs was to publish up-to-date info, the downside is editorial glitches, but I guess we'd rather live with the latter.

PR #1693, opened 2017-08-02
Suggest to remove the term building blocks

ethanport:
Let's stick with flow elements, since the visual icon actually represents an XML element.  The XML element is what's actually defining the application, except with Flow Designer, where the XML is invisible to the developer (but is actually still there).


fermujica:
Thanks Ethan! I agree with you. We should remove the building block term from our glossary. 
Unfortunately, these are not the only docs using this term. 

Let me review the required changes and commit the changes to this branch.

PR #1703, opened 2017-08-02
Update change-target-output-transformation-design-center-task.adoc

Mrk-Nguyen:
Screenshot needs to get updated. "Customize data type" is now "Set data type"


nearnshaw:
What this topic was missing was a mention of how to delete the original default transformation that points to payload.  I just added that in a separate commit.

Otherwise, what you're left with is a Transform component that has two DataWeave scripts in it, one pointing to the target you selected and another pointing to the payload

PR #1704, opened 2017-08-02
Delete add-another-output-transform-design-center-task.adoc

Mrk-Nguyen:
https://github.com/mulesoft/mulesoft-docs/blob/master/design-center/v/1.0/change-target-output-transformation-design-center-task.adoc

talks about outputting to different variables already. We don't need this page.


nearnshaw:
Mark, these two pages are actually covering different things.

The subtle distinction is that what this page describes results in you having a single Transform component that contains TWO DataWeave transformations in it, each sending an output to a different part of the Mule event object.

What the other page covers is changing the output of the existing DataWeave script, it results in you having just one DW script.

Mrk-Nguyen:
I suggest then we change the title of this page to "To Add Another Transformation"

nearnshaw:
Yup, that sounds good.  I'll change the title soon.  Thanks!

PR #1714, opened 2017-08-03
Update dataweave.adoc

Mrk-Nguyen:
I also propose we switch to `(value, index) ->` for our For Each operations and examples (map, mapObject, reduce, pluck, etc..) instead of using $ and $$. I believe this style can reduce confusion for people who are new to DataWeave and reinforces good explicit coding practices.

We can then add a section explaining $ and $$ and tell people that it's a shortcut to writing out the full lambda expression.

### DIFF
--- a/mule-user-guide/v/4.0/dataweave.adoc
+++ b/mule-user-guide/v/4.0/dataweave.adoc
@@ -5,12 +5,14 @@
 
 == Overview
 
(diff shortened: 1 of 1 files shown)


nearnshaw:
Thanks Mark, all of these changes are spot on!!

And I fully agree with your point about $ and $$, I've actually been changing that in a few other places, and we've had similar discussions about that with Ethan Port about the same too.

PR #1718, opened 2017-08-03
Improve Mule 4 change reference

dandiep:
alphabetize, add better descriptions, reorganize into more logical structure.


nearnshaw:
Thanks Dan, this is a lot more readable!

I just fixed the table structure, because there were a number of empty cells appearing in the second table.

PR #1725, opened 2017-08-04
Added clarity for the asset deletion

amitsaxena24:
added notes under the == To Delete an Asset


sleslie23:
Thanks @amitsaxena24 !!!

PR #1733, opened 2017-08-10
Updated pattern matching syntax

leansh:


nearnshaw:
Thanks Shoki!
Adding these changes!

PR #1740, opened 2017-08-16
Update managing-schedules.adoc

lee3star:
cloudhub poll ignores the timezone setting and uses UTC. the current doc only says:

"
Note that all schedules are based in UTC timezone, regardless of what region your CloudHub workers are on.
"

it does NOT clearly mention that the timezone configuration in code is ignored.


george-hoffer:
Thank you Sean!

PR #1742, opened 2017-08-16
Update custom-log-appender.adoc

manibh:
adding * Only Asynchronous log appenders can be used, Synchronous appenders should not be used. to the important section as it will break the functionality if one uses Synchronous appenders.


george-hoffer:
Thank you Mani. If you are ok with this merge, would you delete your branch?  Thanks!

PR #1745, opened 2017-08-16
Remove NT from the documentation

adobni:
Because no one uses it anymore.


nearnshaw:
Thanks Ale, yes that "NT" is redundant and outdated :)

PR #1748, opened 2017-08-17
Typo fixes

george-hoffer:


george-hoffer:
Fer verified and approved changes

PR #1750, opened 2017-08-21
Added else clause in match operator

Mrk-Nguyen:


nearnshaw:
Thanks Mark
I'll add that in to make the example more complete :)

PR #1751, opened 2017-08-21
GA support for COBOL copybook

jfielding:
Removing note for early access support since COBOL copybook support is GA, see Mule 3.8.3 release notes.


george-hoffer:
Thank you Jay!

PR #1752, opened 2017-08-22
Update alerts-on-runtime-manager.adoc

arielmira:
"Application Deleted" Alert for Hybrid deployments is not available, instead "Application undeployed" exists


george-hoffer:
Thank you Ariel - would you click Delete branch if you're satisfied with this merge?  Thank you.

PR #1755, opened 2017-08-23
Improve release notes

afelisatti:
Make sure the HTTP streaming thing is presented as an addition and not a behaviour change.


nearnshaw:
Thanks Ana!

PR #1756, opened 2017-08-24
There is no variable/function named *a*

ethanport:
The module has only the one function myFunc.


nearnshaw:
Thanks Ethan! Yes, that "a" function didn't exist in the example. Merging your changes.

PR #1760, opened 2017-08-25
Update kafka-connector-release-notes.adoc

jpestalardo:
added release notes for v 2.0.1


george-hoffer:
Hi - There are too many spurious characters in this update -- please don't insert items like multiple equal signs. Also only use a table header such as [%header%autowidth.spread] -- the one you used is no longer what we use. 
I entered your changes manually. Verify at https://docs.mulesoft.com/release-notes/kafka-connector-release-notes within 30 mins

PR #1766, opened 2017-08-28
Update database-sync-bulk-insert-task.adoc

Mrk-Nguyen:
Fixed the reason why we need to use syncResult variable.


krishahn:
Tks!

PR #1771, opened 2017-08-30
Update netsuite-connector.adoc

diegosmulesoft:
removed "The connector would then be responsible for mapping SearchRows into the corresponding Record type object in order to facilitate usability"

this caused an issue today with one of our customers and after validating with engineering the connector will always return a map and not a Record type object. We should fix this in all versions of netsuite documentation


george-hoffer:
Thank you Diego

PR #1772, opened 2017-09-01
Updated url to Platform REST api reference

marcelbakker:
Updated url to Platform REST api reference because version was updated and referenced version of RAML seems broken


fermujica:
Thanks! I forgot to update this URL. 
I merged your PR, and I'm deleting this branch.

PR #1781, opened 2017-09-07
Update mq-faq.adoc

adrianmartinmulesoft:
Add FAQ based on a customer support case.


george-hoffer:
Hi Adrian, Thanks for this fix. I needed to make some adjustments to the heading. So fixed manually. Should appear live within 30 mins.

PR #1784, opened 2017-09-11
change case in section heads

ra:


george-hoffer:

We don't capitalize words like "a" in headings, but I agree that MEL and SQL should be. After you complete the contributor agreement, we can add that change. Thank you.

george-hoffer:
Capitalized the use of MEL and SQL in the Database connector headings per your request.

PR #1788, opened 2017-09-13
Fixing typo in the flat file schema doc

jfielding:
Remove double "the" in the note section


george-hoffer:
Thank you Jay. If you're satisfied with this correction, please delete your branch.

PR #1795, opened 2017-09-15
Update managing-applications-on-your-own-servers.adoc

adrianmartinmulesoft:
Clarification on the apps that are already in the on-premise servers or are deployed manually (by dropping the zip in apps folder)


sleslie23:
Thanks @adrianmartinmulesoft !!!

PR #1806, opened 2017-09-23
There is a redundant step in PCE installation

pablodcar:
The Gravity User is created with the same step `curl -X POST http://bandwagon-mulesoft.default.svc/api/complete`


pablodcar:
Tested locally. Gravity user and Anypoint user are created with a single call.

sleslie23:
Thanks @pablodcar !!!

PR #1811, opened 2017-09-26
Update cloudhub-fabric.adoc

jakecmorgan-mule:
added that local persistent queues to the worker are used


jakecmorgan-mule:
added Word Choice Java vs JAVA to README.adoc

PR #1812, opened 2017-09-26
mark CloudHub roles as deprecated

cmordue:
Mark CloudHub Developer and CloudHub Support roles as deprecated since they are no longer created by default for new organizations or business groups.


sleslie23:
Thanks @cmordue !!!

PR #1813, opened 2017-09-27
Redundant page on Proxying

nuggetman:
The page "setting-up-an-api-proxy.adoc" has been decomposed and all of the same content is now available in the remaining "proxy*.adoc" pages. Additionally the page "setting-up-an-api-proxy.adoc" is not in the _toc.adoc whereas the other proxy pages are, therefore this page is redundant.


krishahn:
Thanks for flagging the problem. There are two problems with deleting the page. 1) We would need a redirect and we do not want any more redirects. 2) All the info in setting-up-a-api-proxy is not covered in the subdocs of https://docs.mulesoft.com/api-manager/proxy-about, although there might not be anything critical that isn't covered. I'll remove the duplicated stuff from the page and list the page having the page, now entitled "Proxy Deployment Reference" in the toc.

krishahn:
committed fixes

PR #1820, opened 2017-09-27
Revert "Fixed broken URL reference"

sleslie23:
Reverts mulesoft/mulesoft-docs#1819


sleslie23:
Reverting because submitter has not signed the contributor agreement.

PR #1850, opened 2017-10-03
getting started

juandesi:


mule-sdk/v/4.0/0_getting_started.adoc
```diff
@@ -11,6 +11,24 @@ You must have the following components installed and working in your local envir
 
 * Apache Maven, version 3.3.9 or higher, to manage your project’s build
```
marianogonzalez: Also add the minimum required JDK

juandesi: we only support jdk 8, any jdk 8

mule-sdk/v/4.0/0_getting_started.adoc
```diff
@@ -11,6 +11,24 @@ You must have the following components installed and working in your local envir
...
+If you are using IntelliJ go to: `Preferences > Build, Execution, Deployment > Compiler > Java Compiler`
+and add `-parameters` in the *Additional command line parameters* section.
+
+image::getting_started/intellij_config.gif[600,600]
+
+==== If you use Eclipse
+Eclipse users should right click on the project and then go to: `Java Compiler` and
+enable the *Store information about method parameters (usable via reflection)*
+
+image::getting_started/eclipse_config.gif[600,600]
```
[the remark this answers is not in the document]
juandesi: is studio :p

dukesphere:
I'll await Mariano's ok before merging.

Also, I see there's been a lot of work on the filenames, so I hate to ask, but is there another way we can name the files without prepending chapter and section numbers, such as structure_intro, structure_content_parameters, metadata_intro, metadata_keys, metadata_input, etc. An issue is that once we do updates, and add or remove sections, etc., the filenames are very likely to change. However, we don't want to change filenames once files go to master because customer's bookmarks end up getting 404s once the file name changes, and they complain. We are also trying to get away from having to create redirects after file names change. And finally, we are trying to get away from the idea of our docs as books, but rather as topics that help people understand how to do discrete tasks.

juandesi:
we are prepending numbers so we can keep de docs organized in the local filesystem, there are no really chapters prefixes but those names are the best that we could find to precisely describe whats in the document. Why are the names going to change? maybe we should review that when that happens? 

If you feel that we are doing something wrong let us know and will fix it :). 

Thanks for the feedback

PR #1863, opened 2017-10-05
best practices

marianogonzalez:


mule-sdk/v/4.0/20.0_best_practices_intro.adoc
```diff
@@ -1,2 +1,96 @@
...
+```
+org.mule.salesforce
+    .api.*
+    .internal.*
+```
+Where _api.*_ contains all the classes that are going to be exported and seen by the application (Return Types, Subtypes, Input Types, 
+Connection Client interfaces) and _internal.*_ contains all the classes that contains the specific behaviour of the module 
+(Annotated classes, Connection and Operation implementations).
+
+This way we separate both worlds and get a clear look on what should be exported and what should not.
```
juandesi: add that *ONLY* classes defined in the connector should be exported and not classes from other dependency libs

dukesphere: Awaiting action on change request before merging new content. Thanks!

PR #1867, opened 2017-10-06
isolation

juandesi:


mule-sdk/v/4.0/4.0_isolation_intro.adoc
```diff
@@ -1,6 +1,96 @@
...
+the other one "X" version `2.3`. The `JsonObject` class lives in the same package in both
+versions but the implementation is *very* different between each other, which made them incompatible.
+
+The developers of the modules decided that it was a good idea to expose the `JsonObject`
+class as the return type of their operation:
+
+Module 1
+[source, java]
+----
+public JsonObject getEmployee(@Connection connection, String id) {
```
marianogonzalez: doesn't compile

mule-sdk/v/4.0/4.0_isolation_intro.adoc
```diff
@@ -1,6 +1,96 @@
...
+  return new JsonObject(connection.getById(id));
+}
+----
+
+Module 2
+[source, java]
+----
+public JsonObject getAllProducts(@Connection connection) {
```
marianogonzalez: doesn't compile

marianogonzalez: linenums

mule-sdk/v/4.0/4.0_isolation_intro.adoc
```diff
@@ -1,6 +1,96 @@
...
+We have 2 modules with 2 very different operations, both of them uses the same library "X"
+that parses Json files and returns an instance of a `JsonObject` class. One of them uses "X" version `1.1` and
+the other one "X" version `2.3`. The `JsonObject` class lives in the same package in both
+versions but the implementation is *very* different between each other, which made them incompatible.
+
+The developers of the modules decided that it was a good idea to expose the `JsonObject`
+class as the return type of their operation:
+
+Module 1
+[source, java]
```
marianogonzalez: linenums

mule-sdk/v/4.0/4.0_isolation_intro.adoc
```diff
@@ -1,6 +1,96 @@
...
+
+=== Knowing the Exported classes
+
+It is recommended that *all* exported classes belongs to the module, it is not a
+good practice to return or receive objects from classes that belongs to a library
+that the module is using, because could end up in a class clash.
+
+For example, think the next scenario:
+
+We have 2 modules with 2 very different operations, both of them uses the same library "X"
```
marianogonzalez: can't we just say jackson , jersey or anything real?

mule-sdk/v/4.0/4.0_isolation_intro.adoc
```diff
@@ -1,6 +1,96 @@
...
+Module 2
+[source, java]
+----
+public JsonObject getAllProducts(@Connection connection) {
+  return new JsonObject(connection.getProdsJson());
+}
+----
+
+But now when developing a Mule application, and the Mule Dev tries to use both modules, *one and only one*
+of the two classes implementations will be loaded, since they have exactly the same full qualified name, but not only that!
```
marianogonzalez: ` *one and only one* of the two classes implementations will be loaded`... this is not exactly like that... I wouldn't go into much detail and just say that there will be a conflict which will result in errors

mule-sdk/v/4.0/4.0_isolation_intro.adoc
```diff
@@ -1,6 +1,96 @@
...
+since classloading filtering is done by package all the other classes that lives in the same package
+of that class are going to be exported, resulting in much more complex possible errors.
+
+This will end up in a bunch of `ClassNotFound` and `MethodNotFoundException` between others.
```
marianogonzalez: Again, not so many details. There's a very wide range of errors that you could get here

mule-sdk/v/4.0/4.0_isolation_intro.adoc
```diff
@@ -1,6 +1,96 @@
...
-Connectors can't access libraries provided in the application (like drivers or third party native libraries)
+To introduce the new classloading schema in Mule 4 we need to understand the classloading problems
+that we had in Mule 3.
+
+* Modules were not isolated from each other so is not possible to use different libraries in them.
+* Modules were not isolated from the runtime.
+* Modules were not isolated from the applications.
+
+This means that modules couldn't ship any library version, since any lib which version doesn't match
+with one of other module/application/runtime will end up in not found classes making this a hard problem
```
alegmarra: weird phrasing... "Modules that use a library may conflict with other modules, app or runtime dependencies, causing problems that arise from different versions loaded than the one expected by the module, like ClassNotFoundException or NoSuchMethodException, causing big issues to app developers"

mule-sdk/v/4.0/4.0_isolation_intro.adoc
```diff
@@ -1,6 +1,96 @@
...
+for application and module developers.
+
+So in Mule 4, to tackle this problem, there is a brand new classloading schema that
+makes the runtime, applications and modules isolated from each other. Each one of them
+specifies which *packages* (filtering is made by package and *not* made by class) will
+export and make visible for other modules and applications.
```
alegmarra: "will export as part of their API, making them visible"

mule-sdk/v/4.0/4.0_isolation_intro.adoc
```diff
@@ -1,6 +1,96 @@
...
+
+== Modules Classloading Isolation
+
+As said before, in Mule 4 classloading filtering is done by package and not by class,
+this means that the modules need to specify which packages is going to export for others to
+see. Lucky for you, the SDK makes all this work for you, but it's pretty important that
+you, as a module developer, understands this rule.
+
+It's simple, the way that the SDK knows what packages to export is by introspecting
+each operation and source return and input types and getting its packages. This means
```
alegmarra: each configuration and connection and component

mule-sdk/v/4.0/4.0_isolation_intro.adoc
```diff
@@ -1,6 +1,96 @@
...
+
+Where _api.*_ contains all the classes that are going to be exported and seen by
+the application (Return Types, Subtypes, Input Types, Connection interfaces) and
+ _internal.*_ contains all the classes that contains the specific behaviour of the module
+(Annotated classes, `Connection` and `Operation implementations`; and internal classes).
+
+=== Knowing the Exported classes
+
+It is recommended that *all* exported classes belongs to the module, it is not a
+good practice to return or receive objects from classes that belongs to a library
```
alegmarra: is not a good practice => is *highly* discouraged and _never_ recommended to return...

mule-sdk/v/4.0/4.0_isolation_intro.adoc
```diff
@@ -1,6 +1,96 @@
...
+that the module is using, because could end up in a class clash.
```
alegmarra: `in a class clash` => in a library clash rendering the module incompatible with other modules, app dependencies or even runtimes

mule-sdk/v/4.0/4.0_isolation_intro.adoc
dukesphere: I will wait for action on Mariano's change request before merging pull request.

alegmarra: 2 => two

alegmarra: use the same example but naming it with something real, like Gson, jackson or the json library of your like :)

mule-sdk/v/4.0/4.0_isolation_intro.adoc
alegmarra: review, the error will be different for that example

mule-sdk/v/4.0/4.0_isolation_intro.adoc
```diff
@@ -1,6 +1,96 @@
...
+this means that the modules need to specify which packages is going to export for others to
+see. Lucky for you, the SDK makes all this work for you, but it's pretty important that
+you, as a module developer, understands this rule.
+
+It's simple, the way that the SDK knows what packages to export is by introspecting
+each operation and source return and input types and getting its packages. This means
+that you never want to share a package between an exported class and another class that
+you *don't* want to export.
+
+The suggested package hierarchy (the one that we use for our modules) is:
```
alegmarra: add to the suggested package the namespace of the extension, to avoid "two packages that are equal" in different extensions. So: `sfdc.api` & `sfdc.internal`

mule-sdk/v/4.0/4.0_isolation_intro.adoc
```diff
@@ -1,6 +1,98 @@
...
+
+Module 2
+[source, java, linenums]
+----
+public JsonObject getAllProducts(@Connection ProductsServiceConnection connection) {
+  return new JsonObject(connection.getProdsJson());
+}
+----
+
+But now when developing a mule application, and the mule developer tries to use both modules
```
pablokraan: This part is wrong. You can't have two different plugins exporting the same package.
Unless... there is a dependency against those plugins. For example, mule-file-commons exports some packages that file-connector also exports. This is OK because it will be the same package.
On any other situation, deployment will fail.

You can have classloading issues as the ones you mentioned when two connectors have an operation returning "Object", but internally you know that both are returning some Foo class and you want to pass a Foo from one connector to the other, or try to cast, etc

mule-sdk/v/4.0/4.0_isolation_intro.adoc
```diff
@@ -1,6 +1,98 @@
...
-Connectors can't access libraries provided in the application (like drivers or third party native libraries)
+To introduce the new classloading schema in Mule 4 we need to understand the classloading problems
+that we had in Mule 3.
+
+* Modules were not isolated from each other so is not possible to use different libraries in them.
+* Modules were not isolated from the runtime.
+* Modules were not isolated from the applications.
+
+
+This means that modules that use a library may conflict with other modules, applications or
```
pablokraan: Maybe you need to write this in the past to make it clear that it was the situation on 3.x

mule-sdk/v/4.0/4.0_isolation_intro.adoc
```diff
@@ -1,6 +1,98 @@
...
+
+=== The @Export annotation
+
+For extreme corner or complex cases when for some reason a module requires to export a class
+that is not automatically exported by the SDK, exist the `@Export` annotation.
+
+The `@Export` annotation receives an array of classes to be exported.
+
+[WARNING]
+Take in mind that exporting a class will end up exporting all the classes in the same
```
pablokraan: Maybe too late... but I think this will be confusing for users. If you annotate a class, it will make sense to export that class only, so it makes more sense to me to export a package by name.

pablokraan: Do you want to document privileged API too?

juandesi: yes, maybe we can change this in the future

juandesi: nope

PR #1873, opened 2017-10-09
result y mediaType

marianogonzalez:


mule-sdk/v/4.0/result_object.adoc
```diff
@@ -0,0 +1,70 @@
...
+[source, java, linenums]
+----
+public Result<InputStream, FileAttributes> read(String path) {
+  InputStream content = getContent(path);
+  FileAttributes attributes = getAttributes(path);
+
+  return Result.<InputStream, FileAttributes>builder()
+        .output(content)
+        .attributes(attributes)
+        .mediaType(bestGuessMediaType(path))
```
juandesi: a clear example setting for example an xml would be much better

mule-sdk/v/4.0/result_object.adoc
```diff
@@ -0,0 +1,70 @@
...
+}
+----
+
+== Void attributes
+
+Another possible scenario is that you want to set the payload and the mimeType, but you don't really
+want to set attributes. For that, you use the <Void> type:
+
+[source, java, linenums]
+----
```
juandesi: queres is the void type in this example? maybe clarifying that no attributes = void attributes will be better

PR #1877, opened 2017-10-10
SE-5635 new way to specify date types in apikit for odata

dfmacias:


krishahn:
Can we delete SE-5635 branch?

dfmacias:
yes, sure! thank you

PR #1879, opened 2017-10-10
Update mule-3.9.0-release-notes.adoc

krishahn:
We heard on docs-feedback that Matias B. requested a change from Ubuntu 15.04 to 16.04. Maybe it should be 15.04 and 16.04.


fermujica:
This was already updated in a recent commit.

PR #1882, opened 2017-10-10
DOCS-2302 : Fix FFD file extension and clarify>1 count references (Mule 3.8)

bgroeschen:
In Anypoint Studio 6.4.0 with Mule EE Runtime 3.8.5, DataWeave metadata only finds Flat File Definitions with file extension *.ffd, not *.ffs, so I renamed it. Also, when an invalid value for 'count' is provided, the parser error states, "Value count must be an integer or the string 'unbounded'". Details for 'count' have been updated to mention 'unbounded' as a value equivalent to '>1', and to clarify that integer values should not use quotes (i.e. count: 1 instead of count: '1'). Some of the examples have been changed to reflect these clarifications.


george-hoffer:

george-hoffer:

PR #1884, opened 2017-10-10
DOCS-2301 : Fix FFD extension and mention count: 'unbounded' (Mule 3.9)

bgroeschen:
In Anypoint Studio 6.4.0 with Mule EE Runtime 3.9.0, DataWeave metadata only finds Flat File Definitions with file extension *.ffd, not *.ffs, so I renamed it. Also, when an invalid value for 'count' is provided, the parser error states, "Value count must be an integer or the string 'unbounded'". Details for 'count' have been updated to mention 'unbounded' as a value equivalent to '>1', and to clarify that integer values should not use quotes (i.e. count: 1 instead of count: '1'). Some of the examples have been changed to reflect these clarifications.


george-hoffer:

bgroeschen:
Hi George,

Thanks for the notification. I attempted to follow the link to the MuleSoft
Contributor’s Agreement
<https://api-notebook.anypoint.mulesoft.com/notebooks/#380297ed0e474010ff43>
 form, but all I got was an untitled API notebook. I then clicked Connect
using Github <https://api-notebook.anypoint.mulesoft.com/notebooks/#>, and
then clicked Authorize with GitHub in the pop-up that appeared. However, I
still only see a blank, untitled notebook. What do I do next?

[image: Inline image 1]

Sincerely,
Ben

PR #1898, opened 2017-10-12
Update http-connector-reference.adoc

zaldip:
default response timeout in seconds.


george-hoffer:

george-hoffer:
Thank you Pablo. I found that responseTimeout appears twice in this references, one for the attributes of a request and also for the request builder. I fixed it in both places and in versions 3.8 and 3.9 of this doc.

PR #1901, opened 2017-10-13
Update attachment-transformer-reference.adoc

melamedjoel:
Just taking out a repeated "to" in the documentation


george-hoffer:
Thank you Joel!

PR #1903, opened 2017-10-14
DOCS-2300 : correct name of the plugin mentioned (Mule 3.9)

zdenekca:
I am still new to Mule but this seems to be incorrect based on the pom file example that follows the text. 

On a side note, I am not clear when one would use maven-mule-plugin rather than mule-app-maven-plugin plugin?

thank you


george-hoffer:

PR #1907, opened 2017-10-16
Connector release notes

IanWhyte:


IanWhyte:
Thanks for accepting this. The reason I hadn’t pinged you (that I had fixed the file) is that our staging server was down, so I couldn’t see a build - but it looks fine as published.

Ian

PR #1922, opened 2017-10-23
DOCS-2299 : Update flow-reference-component-reference.adoc (Mule 3.9)

zdenekca:
Clearly state that the referenced flow will take on the message processing before returning back to the "main" flow.


zdenekca:
Or we could say that the referenced flow will be synchronous to the triggering flow unless the Flow Reference Component is wrapped in Async Scope.

george-hoffer:

zdenekca:
I tried to complete the contributor's agreement but I don't see the form. The form link brings me to API Notebook with nothing in it?
_snapshot

zdenekca:
Thank you.

PR #1924, opened 2017-10-23
Adding clarification to upgrade procedures. [DOC REVIEW]

sleslie23:
Creating PR for doc review.

### DIFF
--- a/anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
+++ b/anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
@@ -175,9 +175,9 @@ You must pass the nodename of the node where you are running this command.
 ./manual_update_161.sh finalize-update 
 ----
 
-. Update the health check on the load balancer.
+. Enable port 10248 on your load balancer.
 +
-You must enable port 10248 for the load balancer health check.
+The platform uses this port to perform health checks on the load balancer.
 
 == See Also
 
--- a/anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
+++ b/anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
@@ -111,14 +111,12 @@ From the master node, run the following command:
 ./manual_update_161.sh bootstrap-system
 ----
 
-. Peform the system update
-+
-Perform the following command by logging into each node in your cluster. You must run this command sequentially on each node. Wait until this command completes before running it on the next node.
+. Peform the system update by logging into each node of your cluster and running the following command.
 +
 ----
 ./manual_update_161.sh update-system
 ----
-
++You must run this command sequentially on each node. Wait until this command completes on one node before running it on the next node.
 
 . Bootstrap the RBAC configuration in the cluster
 +
@@ -143,7 +141,7 @@ exit
 
 . Drain each of the nodes in your cluster by running the following command on each node in your cluster. 
 +
-You must pass the nodename for each node.
+You must pass the nodename of the node where you are running this command.
 +
 ----
 ./manual_update_161.sh drain=<node-name>
@@ -153,19 +151,19 @@ Before continuing, ensure that all pods are in `running` or `pending` state. No 
 
 . Make each of the nodes in your cluster is schedulable by running the following command for each node in your cluster. 
 +
-You must pass the nodename for each node.
+You must pass the nodename of the node where you are running this command.
 +
 ----
 ./manual_update_161.sh uncordon=<node-name>
 ----
 
-. From the master node, initiate the application update
+. From the master node, start the application update:
 +
 ----
 ./manual_update_161.sh update-app
 ----
 
-. If you are upgrading a one node installation, fix the LDAP config directory permissions
+. If you are upgrading a one node installation, fix the LDAP config directory permissions by running the following:
 +
 ----
 ./manual_update_161.sh fix-ldap
--- a/anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
+++ b/anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
@@ -129,11 +129,14 @@ From the master node, run the following command:
 . Run the following command to determine the name of each node in your cluster:
 +
 ----
-sudo gravity enter
 kubectl get nodes
+NAME          STATUS    AGE
+172.31.0.25   Ready     19h
+172.31.0.37   Ready     19h
+172.31.0.9    Ready     19h
 ----
 +
-Node names are displayed in the name column. This name corresponds to the internal IP address of the node.
+The node name for each node in your cluster is displayed in the name column. Record these names for use in a later step.
 
 . Exit the gravity shell
 +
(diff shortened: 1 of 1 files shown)


anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
```diff
@@ -28,20 +28,22 @@ $ kubectl
...
-           │     └─25249 grep --color=auto gravitational
-             ├─gravity__gravitational.io__teleport__1.2.0.service
-             ├─gravity__gravitational.io__planet-master__0.1.39-138.service
+$ systemctl status | grep gravitational
+   │     └─25249 grep --color=auto gravitational
+     ├─gravity__gravitational.io__teleport__1.2.0.service
+     ├─gravity__gravitational.io__planet-master__0.1.39-138.service
 ----
++
+You may have to run this command on each node of your cluster. The node that returns the output above is the master node.
```
sleslie23: @sebasampaoli Can you verify this change?

sleslie23:
@prafulrana @rrinaudo FYI, here is the PR for the upgrade doc review. Who else should review this change? Sebas?

sleslie23:
@federicoboerr

anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
anandkumarpatel: Might be clearer to say "the node that Contains planet-master is the master node "

anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
```diff
@@ -188,57 +139,31 @@ kubectl get nodes
...
 ----
 
-. Drain each of the nodes in your cluster.
+. Drain each of the nodes in your cluster by running the following command on each node in your cluster. 
 +
-From the master node, run the following command one each node in your cluster. You must pass the nodename for each node.
+You must pass the nodename of the node where you are running this command.
 +
 ----
 ./manual_update_161.sh drain=<node-name>
```
anandkumarpatel: Can we not get the node name with something like `hostname -i`?

federicoboerr: will not work on every environment.

anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
federicoboerr: Right, good idea Anand.

anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
```diff
@@ -250,9 +175,9 @@ of 'kubectl get nodes' from inside Gravity and wait the node to be 'Ready' again
 ./manual_update_161.sh finalize-update 
 ----
 
-. Update the health check on the load balancer.
+. Enable port 10248 on your load balancer.
```
federicoboerr: This has to be turned into a new section before `Performing the Upgrade`.
It can and should be done prior to the update.

The load balancers sit before the nodes. The load balancer checks an health endpoint of each of the nodes to understand if a node is ready to receive traffic.

## Update load balancer health check
Port `10250` which was used before as a health endpoint is no longer open in `1.6.x and above` and the health of the nodes is now exposed in `HTTP:10248/healthz`.

anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
anandkumarpatel: Then it might be more clear to define how to map node name and what node I am on currently. (You can make write kubectl get nodes --format "node name: () node is ()" ) because at the moment it is not clear how to map the nodes I got from kubectl to the node I am on. 

Minor nit Because I know it's just look at ips but we the easier we can make this experience the better for our user :)

sleslie23: I agree with @anandkumarpatel This is already a long, tedious process. The more clarification we can provide, the better.

anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
sleslie23: @sebasampaoli Can you verify? When creating the original draft we I thought we intentionally added this as the last step..

anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
sleslie23: Updated the docs to say:

You may have to run this command on each node of your cluster. The node that displays `planet-master` is the master node.

anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
sebasampaoli: Just double checked with Fede, @sleslie23 as Fede said we can move it to another section, before the upgrade is performed.

anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
```diff
@@ -188,57 +139,31 @@ kubectl get nodes
...
-pod "pithos-app-install-95fa7b-58flh" evicted
-...
-...
-...
-----
-+
 Before continuing, ensure that all pods are in `running` or `pending` state. No pod should be in `crashloopbackoff` or `terminating` state.
 
-. Make each of the nodes in your cluster is schedulable.
+. Make each of the nodes in your cluster is schedulable by running the following command for each node in your cluster.
```
sebasampaoli: I think here it should be "Make sure" or "Make each of the nodes in your cluster schedulable".

anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
```diff
@@ -57,33 +59,11 @@ sudo ./upload
...
-pod "gravity-site-um5ry" deleted
--> wait until gravity site is running
--> upload the update
 new application anypoint 1.6.1 has been uploaded
 ----
-+
-Depending on your network configuration, this command may take a while to complete. Wait until the command finishes before proceeding to the next step.
 
 
 . Download the installation script and copy it to each node in your cluster.
```
sebasampaoli: We might change "installation script" by "manual upgrade script" or something like that.

anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
sleslie23: @sebasampaoli @federicoboerr 

Should we also change the order in this doc:

https://docs.mulesoft.com/anypoint-private-cloud/v/1.6/install-workflow

anypoint-private-cloud/v/1.6/upgrade-1.6.1.adoc
sleslie23: @sebasampaoli Fixed!

PR #1939, opened 2017-10-27
add missing words "environment ID"

the-jude-abides:
This sentence was missing some words. It seems clear, from context, that it should say:

> In order to create a Mule project, you must enter an API Key and an environment ID.


george-hoffer:
Thank you Jude!

PR #1940, opened 2017-10-27
fixed link to "Create an API key"

the-jude-abides:
It looks like the page this relative link points to has been reorganized from /administration to /security. Fixed link to point to the correct place.


george-hoffer:
Thank you Jude!

PR #1946, opened 2017-11-01
Deprecate populate_repo

adobni:
It has been deprecated and it has been outdated for some time. Configuring the Maven repositories is the current recommended procedure. Remove it for 3.7 and 3.8 too.


fermujica:
Thanks Ale! I update this content for 3.8 and 3.7 as well.

PR #1947, opened 2017-11-02
increase specificity on SAML NameID field

cmordue:
Specifically call out that the NameID field that is used by default for the value of the username is found in the 'Subject' part of the SAML document (as opposed to the AttributeStatements where the rest of the user's attributes are found).


krishahn:
@sleslie23 did u see this?

PR #1961, opened 2017-11-10
reprashing and included headers details

jmjdamen:
It's common to pass client_id and client_secret via headers so included to make it explicit
Changed trait to client-id-required instead of wrong and confusing rate-limited, also changed to RAML 1.0


krishahn:
Making major doc changes in a branch and will incorporate your changes  there. Not merging. Thanks.

krishahn:
Other changes in this PR are made to reorganized docs in https://docs.mulesoft.com/api-manager/prepare-raml-task and https://docs.mulesoft.com/api-manager/client-id-based-policies.

PR #1995, opened 2017-11-17
Adding section for adding external users.

sleslie23:
@aj90 Here is the PR.

### DIFF
--- a/access-management/v/latest/users.adoc
+++ b/access-management/v/latest/users.adoc
@@ -65,7 +65,7 @@ From the External Users tab in Access management, you can view a list of your ex
 
 Each external user on the list must have an anypoint platform acount and they must logged into the public portal at least once. When a user logs into Anypoint Platform for the first time they are automatically added to the External Users tab.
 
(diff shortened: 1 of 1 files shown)


access-management/v/latest/users.adoc
```diff
@@ -59,4 +55,30 @@ By default in every new organization and business group when first created, you
...
+== Inviting an External User to Access an API on a Public API Portal
+
+You can invite users who are not part of your organization to access a public portal. This enables external users to view assets of your public portal and register client applications against a public API portal.
+
+External users are listed in Access Management under the External Users tab of the Users page. This tab lists users from external Anypoint organizations who have registered API client applications against your public API portals.
+
+Before you can invite an external user, the user must:
+
+* Be a registered Anypoint Platform user.
+* Login to your public portal.
```
blai: Shouldn't this be `Login to their own Anypoint Platform account`? (Noticed this is `Before you can invite an external user`)

sleslie23: @

PR #2000, opened 2017-11-17
Update external-oauth-2.0-token-validation-policy.adoc

PabloCamino:
There has been plenty of confusion around what this policy what can do based on its name, so a clear warning seems in place.
For reference: https://mulesoft.slack.com/archives/C02HGAD2E/p1493152384227923


krishahn:
The docs-feedback Slack message you refer to is from April 25. In April, we inserted the warning and made changes in lots of the current docs to try to make this clear. Some of the confusion might be due to the fact that using this policy with the Google provider (I tried it myself a while ago) _seems to work_, and I think there is info in the wild that it works. 

The doc related to this PR already says "To use the OAuth 2.0 Access Token Enforcement Using External Provider policy, you need a Mule OAuth 2.0 provider to provide an access token." I'm adding your sentence, "You can't use this policy to secure your APIs against any other kind of OAuth 2.0 provider, such as Facebook, Google or Azure." To clear up confusion, we should explain what we mean by "you can't use this policy ... " Is there a technical reason or is the reason related to the support contract?

PabloCamino:
Hi @krishahn , the issue is from a support point of view, as customers are mislead to believe they can connect this policy with other providers and that doesn't work, if it does, it's not supported anyway. To use third party providers you already have the External Client Management feature.

krishahn:
Thanks for explaining @PabloCamino, adding your explanation to the docs.

PR #2007, opened 2017-11-19
Update ex2-rest-connect-faq.adoc

NathanNam:


george-hoffer:
This change is to the wrong file. The changes were applied manually to the correct file.

PR #2014, opened 2017-11-26
Added reference to Nov release for REST API

amitsaxena24:
Added reference to Nov release for REST API


george-hoffer:
Thanks Amit

PR #2021, opened 2017-11-28
Update alerts-on-runtime-manager.adoc

adrianpineyro:


sleslie23:
Adrian, see this PR:

https://github.com/mulesoft/mulesoft-docs/pull/2078

There is a wider discussion about rate limits. I included your suggestions there for the dev/PMs to review.

PR #2027, opened 2017-12-01
Removing erroneous entries from release notes

evangelina:
Removing 3 Jiras that were reported on 3.9.0 SNAPSHOT and therefore don't apply here.


george-hoffer:
Thank you for handling this

PR #2028, opened 2017-12-04
Add reference to the Developer Portal.

ethanport:
This link does not show up in top Google searches, so it will be good to sprinkle this link around the docs. It's an essential management tool.


sleslie23:
@ethanport We're in the process of rewriting all of the getting started content.

PR #2034, opened 2017-12-06
Updated AES Max Size

PabloCamino:
There is a minSize for AES which seems to be 16 byte. So it means by default you are required min 128 bit key. But it can handle 192 and 256 key sizes as per the AES standard, adapting based on key size. It might even accept larger values, so I'd recommend to double check with engineering. But 16 is definitely not the maximum.
<https://mulesoft.slack.com/archives/C02GKLLQ1/p1434138870000857>


fermujica:
Thanks! Yes. per the code, 32 bytes seems correct as the max size. 
Thanks for including your conversation with ENG for reference.

PR #2035, opened 2017-12-07
AOP-1590 Adding 1.7 migration procedures.

sleslie23:


anypoint-private-cloud/v/1.7/upgrade.adoc
```diff
@@ -1,3 +1,98 @@
...
+synchronizing anypoint:1.7.0-rc.3 with Docker registry at 172.31.0.26:5000
+synchronizing anypoint:1.7.0-rc.3 with Docker registry at 172.31.0.45:5000
+synchronizing anypoint:1.7.0-rc.3 with Docker registry at 172.31.0.5:5000
+anypoint:1.7.0-rc.3 has been uploaded
+updating anypoint from 1.6.1 to 1.7.0-rc.3
+update operation (bae00abf-98a7-4ca9-b0d2-06f3bfea3024) has been started
+the cluster is updating in background
+----
++
+Depending on your network configuration, this command may take a while to complete. Wait until the command finishes before proceeding to the next step.
```
a-palchikov: The command will finish with launching the upgrade in the background - the upgrade itself runs inside a systemd unit. That said, the command will finish before the actual update will - worth noting to the user.
Either the `gravity status` or `gravity plan` should then be used to monitor the operation progress.

anypoint-private-cloud/v/1.7/upgrade.adoc
```diff
@@ -1,3 +1,98 @@
...
+
+. Verify that the migration was successful by performing one of the following:
```
a-palchikov: Success of the upgrade can be verified with `gravity status` by noting the cluster state (should be `active`) and application version (should be the current from the upgrade package).

anypoint-private-cloud/v/1.7/upgrade.adoc
```diff
@@ -1,3 +1,98 @@
...
++
+* Run the following command from the node where you performed the upgrade:
++
+----
+$ sudo gravity enter
+$ kubectl get pods
```
a-palchikov: This might also show some Pods in a state different from `Running` and `Completed` which does not give the proper impression of the outcome.

PR #2041, opened 2017-12-11
fixing typo 'environmentis'

jmporcel:


george-hoffer:
Thank you Juan Martin

PR #2042, opened 2017-12-11
replace "it's" with "its"

the-jude-abides:
The correct possessive for "it" should be "its".  "It's" is a contraction for "it is".


george-hoffer:
Thank you Jude

PR #2043, opened 2017-12-11
Corrected Mule service flow.

gramone:
The example provided was incorrect as the service flow was listening on the wrong queue, causing the flow to wait indefinitely.


george-hoffer:
Thank you Gilles. Does this apply only to the Mule 3.9 version of this transport or also to earlier versions and if so which?

gramone:
Hi George,

No problem. It is independent of the Mule version so the correction applies
to all.
Thanks!

Gilles.

PR #2046, opened 2017-12-12
Add new action under Access Management - Users

santro92:
Added a new action called `Password changed` which is executed when the user actually changes the password and is different from `Password reset` which is executed on just the click of the `Reset Password` link.


santro92:
@cmordue Sure I will change it in code and the docs.

PR #2048, opened 2017-12-12
Create mule-3.8.6-release-notes.adoc

anosenzo:


dukesphere:
Adding RN draft on 12-12 in prep for Friday, 12-15 release. It will get updated before the release, and we'll add a link to it at that time. Tested locally and it does not break the build.

PR #2070, opened 2017-12-20
Update cloudhub-deployment-mmp-reference.adoc

WilliamPinho:


mule-user-guide/v/4.0/cloudhub-deployment-mmp-reference.adoc
```diff
@@ -25,8 +27,12 @@ For example:
 
 <properties> +
   <http.port>8081</http.port> +
-</properites>
-
+</properties>
+| workers | The number of workers. By default, it is set to 1.
+| workerType | Size of each worker. By default, it is set to "Medium".
```
fermujica: @WilliamPinho: Is it safe to assume that these are the supported values (and their vCore equivalents)?


> MICRO (0.1 vCores)
> SMALL (0.2 vCores)
> MEDIUM (1 vCore )
> LARGE (2 vCores)
> XLARGE (4 vCores)
>

WilliamPinho: Apart from these values, there are also 'xxLarge' -> 8v cores and '4xLarge' -> 16v cores. However, it is not safe to assume since the tests to validate that are about to be created.

PR #2084, opened 2017-12-29
Explain how Scheduler works in the debugger now.

ethanport:
It's different from the Polling scope in Mule 3.


fermujica:
Thanks Ethan! I applied a few editorial to your content. Feel free to let me know what you think.

PR #2091, opened 2018-01-02
Add example object to User Properties.

ethanport:


dukesphere:
Hey Marra, The content in this file comes from source, so it gets overwritten each time we update it. So any changes to it should take place in the source and not here.

PR #2107, opened 2018-01-06
 Update filters.adoc

estesolutions:
Global Filters description was mistakenly put into the section below it.


estesolutions:
A more thorough restructuring needed. Will be adding to this.

PR #2133, opened 2018-01-19
MULE-15211 : Clarify how the default object store behaves (Connector)

ethanport:
The parameter is actually `objectStore`. Using `Object Store` will fail. When deployed to CloudHub worker(s), the default persistent object store is used, and key/values appear in Runtime Manager in the Mule application's Application Data tab. Properties are case sensitive, so the docs should show the exact correct name.

### DIFF
--- a/connectors/v/latest/object-store-connector-reference.adoc
+++ b/connectors/v/latest/object-store-connector-reference.adoc
@@ -1,5 +1,5 @@
 :toc:               left
-:toc-title:         ObjectStore Connector
+:toc-title:         Object Store Connector
 :toclevels:         2
 :last-update-label!:
 :docinfo:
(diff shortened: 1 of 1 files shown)


dukesphere:
Mariano, This change will get overwritten next time the ref gets updated. So any necessary changes should go into the source and not into this adoc output file.

marianogonzalez:
@juandesi I checked the objectStore parameter on the contains operation does not have a space nor it's capitalized. Is this a bug with the doc generator?

juandesi:
@ethanport @dukesphere @marianogonzalez that works as expected, the name of the parameter is defined to match the UI, not the XML. This will happen for all parameters, and should remain that way. 

Maybe a new "dsl name" element should be added to the table if the information still important, but want to show as many DSL references as we can.

PR #2134, opened 2018-01-19
[SE-7022] - Updated log4j2 example for mule agent logs redirection.

fpiagent:
The log4j2.xml example in these pages was not working for Support nor customers:
- File location was not absolute (so depending on how you start the Mule, the location may change)
- Appender-ref was missing for the new appender.


fermujica:
Thanks for updating the logger configuration Fede. 
I just merged this.

PR #2170, opened 2018-01-29
minor changes on the portal wording

amitsaxena24:
please use lowercase for public portals as this is not an official name yet. Just a context for using them. Thank you.


george-hoffer:
Made changes manually.

PR #2212, opened 2018-02-08
DOCS-2296 : fixes example of using remove on an array (Mule 3.9)

ryan-slack-big:


george-hoffer:

PR #2227, opened 2018-02-15
Update to README.adoc

karna12:
Few corrections

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -7,15 +7,15 @@ ifdef::env-github[]
 :warning-caption: :warning:
 endif::[]
 
-This is the repository for MuleSoft's documentation and contains the source content for the link:https://docs.mulesoft.com/[docs site].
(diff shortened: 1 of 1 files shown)


george-hoffer:

Also, in our asciidoc syntax, we make references with link:/<section>, so putting a space after link: is incorrect.

george-hoffer:
Regarding these suggestions: 
Line 10 can't be added because asciidoc doesn't support a space between link: and the link text.
Line 16 can't be added because we use the Oxford comma style: a, b, and c
Line 18 can't be added because it breaks the link: command.
Line 26 breaks the link: command.
Line 31 is unneeded because asciidoc automatically changes multiple spaces to a single space.
Line 45 and line 90 the fix is incorrect grammar. 
Line 135 is ok.
Line 159 is incorrect grammar.
Line 188 is ok.
Line 194 is ok.

PR #2235, opened 2018-02-15
Documented installing custom Mule modules into Studio

ethanport:


fermujica:
Thanks Ethan, I cherry picked these commits into #2680

PR #2239, opened 2018-02-16
Xml

marianogonzalez:


dukesphere:
FYI: The /migration dir no longer exists. The /migration files should go under /mule-user-guide/v/4.1 now. I will push these in, then move those and add them to the appropriate tocs.

PR #2253, opened 2018-02-20
Update about-connector-certification-program-guidelines.adoc

NathanNam:
Removed the "See Also" part temporarily.


george-hoffer:
Handled manually

PR #2255, opened 2018-02-20
Add Endpoints for Every Region

nishantkakar:


nishantkakar:
Do not merge before 24 Feb 2018

PR #2256, opened 2018-02-21
Adds ClassValue annotation and fixes format for allowInlineDefinition and allowReferences

ndinu:


dukesphere:
I will copyedit after merging in the PR.

PR #2296, opened 2018-02-28
Update sap-concur-connector.adoc

NathanNam:


george-hoffer:
Section removed before PR.

PR #2303, opened 2018-03-01
Update box-connector.adoc

diegosmulesoft:


george-hoffer:
Thanks Diego. Does this fix also apply to earlier Mule versions of this connector?

PR #2306, opened 2018-03-01
Update mule-maven-plugin.adoc

manibh:
Please apply the same change to 3.7.x and 3.8.x and 3.9.x


fermujica:
Thanks for pointing this out. 
This actually happens only if you have other properties defined in your mule-maven-plugin configuration. 
MMP configs override mule-app.properties config. I'm adding this to the docs. 

Thanks again!

PR #2308, opened 2018-03-02
MULE-14594: Document raise-error component

afelisatti:


mule-user-guide/v/4.0/raise-error-component-reference.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+The following example flow produces a `CONNECTIVITY` error and will result in a
+500 response with body `Could not connect to backend.`.
+
+.Example: XML Configuration of the Flow
+[source,xml,linenums]
+----
+<flow name="raise-error-example-flow">
+  <http:listener config-ref="HTTP_Listener_Configuration" path="/"/>
+  <raise-error type="CONNECTIVITY" description="Could not connect to backend."/>
```
marianogonzalez: this example is confusing for the user. No app will do this in real life. 

I think we need to split the documentation of this feature in two:

1) This article should cover the base concept and show examples which make business sense. For example, raising an error when trying to withdraw money from an empty account. It's a business error
2) examples like this one belong on the smart connector documentation

mule-user-guide/v/4.0/raise-error-component-reference.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+| Description | String or DataWeave expression | Specifies the error description. |
+`description="Invalid email. Cannot complete transaction."`
+
+| Type | String | Specifies the type of the error. |
+`type="ORDER:INVALID_DATA"`
+
+|===
+
+Note that you can only raise core errors, such as `SECURITY`, or custom ones.
```
marianogonzalez: what about the namespace of the error? Any limitations there? I remembered we discussed having namespaceless error and the mythical APP namespace, but I got old and forgot what we decided.

afelisatti: Below I mention custom error must feature a namespace that's not present.

mule-user-guide/v/4.0/raise-error-component-reference.adoc
afelisatti: Agreed. Perhaps the first example should be handling another non business error and raising a custom one?

mule-user-guide/v/4.0/raise-error-component-reference.adoc
marianogonzalez: Yes, but it's confusing, because you say that after you first presented the user with a namespaceless error. I think that the rules about namespacing should be explained before examples are presented.

marianogonzalez: Also it'd be good to link to the error handling article for reference on what errors and namespaces actually are

afelisatti: The link is in the beginning when I first mention errors.

afelisatti:
@dukesphere, this is ready for your review now.

PR #2310, opened 2018-03-02
Update microsoft-dynamics-365-release-notes.adoc

jpestalardo:
MS Dynamics 365 1.0.1 version released


jpestalardo:
@george-hoffer  please don't merge it as we have issues with the version

PR #2314, opened 2018-03-05
Correcting mis-spelling

rogerkohler:
Typo error corrected, changed "de" to "the".

Original said - "Use the undeploy goal to undeploy de applications and stop Mule server."


george-hoffer:
Thank you!

PR #2315, opened 2018-03-05
Adds objectstore migration guide documentation

ndinu:


mule-user-guide/v/4.0/migration-connectors-objectstore.adoc
```diff
@@ -0,0 +1,167 @@
...
+</spring:beans>
+
+<objectstore:config name="ObjectStoreConnector" objectStore-ref="myCustomObjectStore"  partition="users" entryTtl="3600000" expirationInterval="10000" maxEntries="1000" persistent="true"/>
+----
+
+On a flow you could use :
+
+[source,xml,linenums]
+----
+<objectstore:contains config-ref="ObjectStoreConnector" key="#[flowVars.userId]"/>
```
alegmarra: merge this snippet with the one above, and change this line with a more real-case scenario, something simple like this https://docs.mulesoft.com/mule-user-guide/v/3.8/object-store-connector#example-use-case `/store` `/retrieve` flows (you can even simplify them further, the idea is to have http, os and maybe weave if necessary) and use the 1, 2, 3.. to explain what is going on. Then, show the exact same scenario in mule 4.

mule-user-guide/v/4.0/migration-connectors-objectstore.adoc
```diff
@@ -0,0 +1,167 @@
...
+[source,xml,linenums]
+----
+<objectstore:store config-ref="ObjectStoreConnector" key="#[flowVars.userId]" value-ref="#[flowVars.userAddress]" overwrite="true" />
+----
+
+.Mule 4 example
+[source,xml,linenums]
+----
+<os:store key="#[vars.userId]" objectStore="customObjectStore">
+			<os:value ><![CDATA[#[vars.userAddress]]]></os:value>
```
alegmarra: use an example were in mule 3 you needed to use a `transform` before storing the value, but now you can do it inline. Something simple like using `payload.username` and `payload.token` instead of storing all the payload

alegmarra: mention that since in mule 4 the transformation is done inline, the payload after invoking the `store` is the same as the one previous to the operation. In mule 3 the transform would've altered the payload for all the flow, and if the original payload was needed after invoking `store`, then you should've save it in a flowvar and restored it later.

mule-user-guide/v/4.0/migration-connectors-objectstore.adoc
```diff
@@ -0,0 +1,167 @@
...
+[source,xml,linenums]
+----
+<os:object-store name="customObjectStore" persistent="false"/>
+----
+
+[[dispose-clear]]
+=== Cleaning an Object Store
+
+The dispose operation on Mule 3 was used to clear a specific partition of an object store.
+On Mule 4 we have the clear operation that clears a whole object store. There is no
```
alegmarra: use bold for the `clears a whole OS`

alegmarra: actually, a `WARNING:` would be even better

mule-user-guide/v/4.0/migration-connectors-objectstore.adoc
```diff
@@ -0,0 +1,264 @@
...
+
+<1> Creates the custom object store using Spring beans and a reference to a java class.
+
+<2> References the custom object store in a configuration.
+
+<3> Before using the store operation, a DataWeave transformation is used to set on the
+payload what is going to be stored.
+
+<4> The payload is stored in the object store.
+
```
alegmarra: Implementing the same app in Mule 4 would look like this:

mule-user-guide/v/4.0/migration-connectors-objectstore.adoc
```diff
@@ -0,0 +1,264 @@
...
+<http:listener-config name="HTTP_Listener_config">
+  <http:listener-connection host="localhost" port="8080" />
+</http:listener-config>
+
+<os:object-store name="booksObjectStore" maxEntries="1000" entryTtl="1" entryTtlUnit="HOURS" expirationIntervalUnit="SECONDS" expirationInterval="10"/> // <1>
+
+<flow name="storeBook">
+  <http:listener config-ref="HTTP_Listener_config" path="/store"/>
+  <os:store key="#[attributes.queryParams.id]" objectStore="booksObjectStore" failIfPresent="true"> // <2>
+    <os:value ><![CDATA[#[%dw 2.0
```
alegmarra: Remove the CDATA and dw 2.0 reference. 
Also, in 3.x snippet you are storing an app/java type (which is fine), so it's not an application/json transformation here. This simplifies the snippet further, you can do:

\<os:value\>#[{
  id: attributes.queryParams.id,
  book: attributes.queryParams.book,
  author: attributes.queryParams.author
}]\</os:value\>

mule-user-guide/v/4.0/migration-connectors-objectstore.adoc
```diff
@@ -0,0 +1,264 @@
...
+output application/json
+---
+{
+  id: attributes.queryParams.id,
+  book: attributes.queryParams.book,
+  author: attributes.queryParams.author
+}]]]></os:value>
+  </os:store>
```
alegmarra: missing `key` parameter

alegmarra: shouldn't `failIfPresent` be true so the example has the exact same behaviour as the mule 3 snippet?

mule-user-guide/v/4.0/migration-connectors-objectstore.adoc
```diff
@@ -0,0 +1,264 @@
...
+</flow>
+
+<flow name="retrieveBook">
+  <http:listener config-ref="HTTP_Listener_config" path="/retrieve"/>
+  <os:retrieve key="#[attributes.queryParams.id]" objectStore="booksObjectStore"/>
```
alegmarra: stored java, so now `<set-payload value="#[output application/json --- payload]">`

mule-user-guide/v/4.0/migration-connectors-objectstore.adoc
```diff
@@ -0,0 +1,264 @@
...
+</flow>
+----
+
```
alegmarra: As you may see, main differences are:

mule-user-guide/v/4.0/migration-connectors-objectstore.adoc
```diff
@@ -0,0 +1,264 @@
...
+<http:listener-config name="HTTP_Listener_config">
+  <http:listener-connection host="localhost" port="8080" />
+</http:listener-config>
+
+<os:object-store name="Object_store" persistent="false"/>
+
+<flow name="storeClient">
+  <http:listener config-ref="HTTP_Listener_config" path="/put"/>
+  <os:store key="#[payload.id]" objectStore="Object_store" failIfPresent="true">
+    <os:value ><![CDATA[#[%dw 2.0 // <1>
```
alegmarra: same script cleanup as before. Note that again formats are different between snippets

PR #2318, opened 2018-03-06
missing steps when deploying assets to exchange

lee3star:
1st, Org ID needs to be set as project's groupId
2nd, type needs to be specified.


george-hoffer:
Thank you Sean!

PR #2326, opened 2018-03-08
explained DW vars vs. functions vs. lambdas

ethanport:


dukesphere:
I'll await merging PR until I hear back from Ethan on this suggestion from MDA:

PR #2332, opened 2018-03-10
Attempted to expand the description of watermark

ethanport:


connectors/v/latest/database-documentation.adoc
```diff
@@ -783,7 +783,37 @@ Updates data in a database.
...
 
 +++
-Selects from a table at a regular interval and generates one message per each obtained row. <p> Optionally, watermark and id columns can be provided. If a watermark column is provided, the values taken from that column will be used to filter the contents of the next poll, so that only rows with a greater watermark value are returned. If an id column is provided, this component will automatically make sure that the same row is not picked twice by concurrent polls
+In Anypoint Studio this operation is called On Table Row. 
+
+Unlike all other Database connector components, you add a Database listner to the source of a flow. The Database listener periodically selects from a database table at a regular interval and generates one message per each obtained row. 
+
+==== Avoiding Duplicate Processing of Database Records
+
+<p> Optionally, a watermark column or id column can be provided to the Database listener configuration. Either one of these column values will be persisted in an object store. You can modify the object store in the
```
marianogonzalez: incomplete and inaccurate sentence. The object store is private

connectors/v/latest/database-documentation.adoc
```diff
@@ -783,7 +783,37 @@ Updates data in a database.
...
+<p> Optionally, a watermark column or id column can be provided to the Database listener configuration. Either one of these column values will be persisted in an object store. You can modify the object store in the 
+
+==== Using an Id Column
+If the table has key or identity columns, where the database quarantees every value is unique, you can assign one of these columns as an *id* column, to prevent duplicate processing of rows from the table. 
+
+When an Id column is provided, this component will automatically make sure that the same row is not picked twice by concurrent polls. In other words, the Id column behaves as a type of cluster-friendly idempotent message filter. 
+
+
+==== Using a Watermark Column
+An Id column has unique values, guaranteeing that every new row will be returned once by a subsequent poll by the database listener.
```
marianogonzalez: this is wrong and contradicts the definition above

connectors/v/latest/database-documentation.adoc
```diff
@@ -783,7 +783,37 @@ Updates data in a database.
...
+==== Using an Id Column
+If the table has key or identity columns, where the database quarantees every value is unique, you can assign one of these columns as an *id* column, to prevent duplicate processing of rows from the table. 
+
+When an Id column is provided, this component will automatically make sure that the same row is not picked twice by concurrent polls. In other words, the Id column behaves as a type of cluster-friendly idempotent message filter. 
+
+
+==== Using a Watermark Column
+An Id column has unique values, guaranteeing that every new row will be returned once by a subsequent poll by the database listener. 
+
+An alternative filtering option is to use a *watermark* column.
```
marianogonzalez: not alternative, they  can be used at the same time. There're even use cases in which you would use both features at the same time, both pointing to the same column (e.g: a self increment primary key)

connectors/v/latest/database-documentation.adoc
```diff
@@ -783,7 +783,37 @@ Updates data in a database.
...
+
+When an Id column is provided, this component will automatically make sure that the same row is not picked twice by concurrent polls. In other words, the Id column behaves as a type of cluster-friendly idempotent message filter. 
+
+
+==== Using a Watermark Column
+An Id column has unique values, guaranteeing that every new row will be returned once by a subsequent poll by the database listener. 
+
+An alternative filtering option is to use a *watermark* column. 
+
+When a watermark column is provided, the values returned from that database table's column are tested against the largest previous value, and rows are only returned if the watermark column is strictly greater than all previous values.
```
marianogonzalez: largest -> greatest

connectors/v/latest/database-documentation.adoc
```diff
@@ -783,7 +783,37 @@ Updates data in a database.
...
+When a watermark column is provided, the values returned from that database table's column are tested against the largest previous value, and rows are only returned if the watermark column is strictly greater than all previous values. 
+
+If the Watermark column is always increasing in the database, then a Watermark column will guarantee that older rows are never returned a second time in subsequent polls, but every new row is returned one time in one of the polling cycles. If a new row is equal to the existing watermark, then the new row is filtered out and not returned.  
+
+Unlike an id column, a watermark columnn can be any database table column,
+but has the additional restriction that the values must be strickly increasing in order to avoid duplicate processing of rows.  
+
+
+Note: 
+Although not a typical use case, you could assign the same identity column to a watermark, in which case only the results will only include rows whose id column is greater than any previous id column.
```
marianogonzalez: super typical use case

connectors/v/latest/database-documentation.adoc
```diff
@@ -783,7 +783,37 @@ Updates data in a database.
...
+
+When a watermark column is provided, the values returned from that database table's column are tested against the largest previous value, and rows are only returned if the watermark column is strictly greater than all previous values. 
+
+If the Watermark column is always increasing in the database, then a Watermark column will guarantee that older rows are never returned a second time in subsequent polls, but every new row is returned one time in one of the polling cycles. If a new row is equal to the existing watermark, then the new row is filtered out and not returned.  
+
+Unlike an id column, a watermark columnn can be any database table column,
+but has the additional restriction that the values must be strickly increasing in order to avoid duplicate processing of rows.  
+
+
+Note:
```
marianogonzalez: Note -> `[TIP]`

connectors/v/latest/database-documentation.adoc
```diff
@@ -783,7 +783,26 @@ Updates data in a database.
 `<db:listener>`
 
 +++
-Selects from a table at a regular interval and generates one message per each obtained row. <p> Optionally, watermark and id columns can be provided. If a watermark column is provided, the values taken from that column will be used to filter the contents of the next poll, so that only rows with a greater watermark value are returned. If an id column is provided, this component will automatically make sure that the same row is not picked twice by concurrent polls
+In Anypoint Studio this operation is called On Table Row. 
+
+Unlike all other Database connector components, you add a Database listner to the source of a flow. The Database listener periodically selects from a database table at a regular interval and generates one message per each obtained row.
```
marianogonzalez: listner -> listener

marianogonzalez: `you add a Database listner to the source of a flow` -> this wording is weird and not really aligned with Mule lingo. There's no `source of a flow`, but rather flows that have message sources. I also think that marketing is going with the term `trigger` for these sort of simplified message sources

connectors/v/latest/database-documentation.adoc
```diff
@@ -783,7 +783,26 @@ Updates data in a database.
...
+In Anypoint Studio this operation is called On Table Row. 
+
+Unlike all other Database connector components, you add a Database listner to the source of a flow. The Database listener periodically selects from a database table at a regular interval and generates one message per each obtained row. 
+
+==== Avoiding Duplicate Processing of Database Records
+
+<p> The Database connector can be configured with various criteria to filter the results returned from a database table. The two filtering options are to set an Id Column or a Watermark column in the database table to be used for filtering rows returned from the database. You can also set both of these filter columns to provide even more sensitive filtering conditions.
```
marianogonzalez: Not the database connector but the trigger

marianogonzalez: id column is not for filtering. Is for idempotency. Two different things

marianogonzalez: `provide even more sensitive filtering conditions.` -> doesn't make much sense

connectors/v/latest/database-documentation.adoc
```diff
@@ -783,7 +783,26 @@ Updates data in a database.
...
+In Anypoint Studio this operation is called On Table Row. 
+
+Unlike all other Database connector components, you add a Database listner to the source of a flow. The Database listener periodically selects from a database table at a regular interval and generates one message per each obtained row. 
+
+==== Avoiding Duplicate Processing of Database Records
+
+<p> The Database connector can be configured with various criteria to filter the results returned from a database table. The two filtering options are to set an Id Column or a Watermark column in the database table to be used for filtering rows returned from the database. You can also set both of these filter columns to provide even more sensitive filtering conditions. 
+
+Both of these filters are reset each time from the rows returned from the next database table polling cycle, to continuously avoid duplicate processing after each database polling cycle.
```
marianogonzalez: `are reset each time from the rows returned from the next database table polling cycle`. not true. Watermarking requires propagation of state

connectors/v/latest/database-documentation.adoc
```diff
@@ -783,7 +783,26 @@ Updates data in a database.
...
+
+==== Avoiding Duplicate Processing of Database Records
+
+<p> The Database connector can be configured with various criteria to filter the results returned from a database table. The two filtering options are to set an Id Column or a Watermark column in the database table to be used for filtering rows returned from the database. You can also set both of these filter columns to provide even more sensitive filtering conditions. 
+
+Both of these filters are reset each time from the rows returned from the next database table polling cycle, to continuously avoid duplicate processing after each database polling cycle.  
+
+==== Using a Watermark Column
+
+When a watermark column is provided, the watermark is automatically set to the maximum value returned from the database table. This sets a lower bound for future database polling cycles. In future polling cycles, the only rows returned are new rows with values in the watermark column strictly greater than the previous watermark.
```
marianogonzalez: best watermark summary i read in a long time 👍

connectors/v/latest/database-documentation.adoc
ethanport: Unlike all other Database connector components, you add a Database listener as a trigger to the beginning of a flow. The Database listener periodically selects from a database table at a regular interval and generates one message per each obtained row.

connectors/v/latest/database-documentation.adoc
ethanport: You can also set both of these filter columns at the same time, as each filter works in a different way.

connectors/v/latest/database-documentation.adoc
ethanport: Both of these filters retain state that is updated from the rows returned from the next database table polling cycle, to continuously avoid duplicate processing after each database polling cycle.

connectors/v/latest/database-documentation.adoc
ethanport: More specific. How's this? ```<p> The Database connector can be configured with various criteria to limit or filter the results returned from a database table. The two options are to set an Id Column or a Watermark column in the database table to be used to limit the rows returned from the database based on certain conditions. 

A *Watermark* column set's a value in the SQL WHERE clause to filter the results returned from the database. 

An *Id Column* provides idempontent criteria to avoid duplicate processing of any row where the Id Column value matches any previously returned value. 

You can also set both of these filter columns at the same time, as each filter works in a different way.```

dukesphere:
@marianogonzalez @ethanport These changes would need to be put in the source code comments for the database connector. Otherwise, they will get overwritten the next time the db connector ref is generated from source.

PR #2334, opened 2018-03-11
MULE-14690: Add attachments info regarding migration

afelisatti:


afelisatti:
Need @juandesi to check whether something could be added around Email and WS consumer.

PR #2340, opened 2018-03-12
Update edifact-module.adoc

adrianmartinmulesoft:
Clarification about OOTB schema location. Support case https://na29.salesforce.com/50034000011lYma


george-hoffer:
Thank you Adrian

PR #2341, opened 2018-03-12
Create Template instead of Transaction

gramone:
Previous instruction made it impossible to reuse transaction as a buyer.


george-hoffer:
Thank you Gilles

PR #2358, opened 2018-03-15
Add Regex example with match flag and match pattern

jamesontriplett:
Adding an extra example around doing regex.  I found a couple of challenging spots that I think this will help.


dukesphere:
thanks, ana

PR #2360, opened 2018-03-15
Update sending-data-from-arm-to-external-analytics-software.adoc

StephenFishman:
this article is valid for on-prem pce customers but the title implies it's valid only for hybrid


StephenFishman:
made a change to the title to indicate that the article was valid for on-prem customers as well as hybrid

sleslie23:
To avoid confusion, the name we have been asked to use is Hybrid.

PR #2363, opened 2018-03-15
Defined "what are mule apps"

ethanport:
for both Studio and flow designer


dukesphere:
I'm manually rolling this into the b-master branch instead of beta-anypt.

PR #2364, opened 2018-03-15
Change toUpper to more interesting example

ethanport:


dukesphere:
I manually rolled this into b-master instead of beta-anypt (no longer used). Merging into b-master not possible.

PR #2378, opened 2018-03-19
Portal customization - Banner height specification

damiancalabresi:
The Exchange Portal banner is going to be fixed to 220px. Exchange team confirmed this information. I wanted to add this to the documentation as it was important for one customer.


george-hoffer:
Thank you Damian

PR #2391, opened 2018-03-22
Update header-inject-remove-task.adoc

FedeAmdan:


krishahn:
Already took care of removing this Fede.

PR #2406, opened 2018-03-24
Update APAC region domains

sundippatel:


nishantkakar:
We're not planning to do Sydney right now.

PR #2421, opened 2018-03-27
schema file locations cannot by dynamic

lee3star:
connector UI shows the schema locations can be 'From Expression', which misleads customers to use MEL with flow variables.

I checked with Dennis Sosnoski, the main contributor of X12 connector project, and confirmed that "it is expecting the values when initialized. There's substantial overhead in loading the schemas, so that's only done at initialization." as a result, MEL with flow variables won't work. In other words, dynamically loading different schema files at runtime is not currently supported.


george-hoffer:
Made change manually due to conflict. Thank you for providing this.

PR #2453, opened 2018-04-04
Checking the DLB health

ethanport:
This needs to be merged with the previous PR.


fermujica:
Thanks @ethanport! I need to go through all this content along with #2451. The DLB docs are being restructured so we can have clearer and more "to-the-point" information instead of a long wall of text. 
I'll definitely incorporate your additions.

PR #2465, opened 2018-04-05
tried to fix broken HTML tags displaying in docs

ethanport:


sleslie23:
Thanks @ethanport !!!

PR #2477, opened 2018-04-06
Clarified mergeWith behavior and description.

ethanport:


dukesphere:
Okay, I did a PR with a change that will add your update to this file once it's incorporated into master: https://github.com/mulesoft/data-weave/blob/master/wlang/src/main/resources/dw/core/Objects.dwl

PR #2493, opened 2018-04-12
Update dw-core-types.adoc for typos.

dhimate:
Fixed typos.


dukesphere:
Manually adding to DW source file in Core.dwl. This will get rolled in with other changes before end of June.

PR #2495, opened 2018-04-12
DMT

juandesi:


dukesphere:
I will copyedit this post-merge via https://www.mulesoft.org/jira/browse/DOCS-2121

PR #2499, opened 2018-04-15
fix missing ]s + remove all the weird quote chars

nathan-at-mulesoft:
bunch of the examples were missing the end square bracket, also the quotes in the examples were not the standard ' character which would create issues copying and pasting..


fermujica:
This was fixed in PR #2510

PR #2525, opened 2018-04-20
Small typo on the java - case of customAssertion

nathan-at-mulesoft:


fermujica:
There was a mismatch between the class name, and the package called from the pom config.
Good catch. Thanks!

PR #2540, opened 2018-04-24
Explained limitations of Mapping vs. Module files

ethanport:
DW Mapping and Module files are similar, but with some critical differences that are hard to troubleshoot. Important to explain clearly in the docs.


dukesphere:
manually "merged" in this change to new mule4-user-guide/v/4.1 dir.

PR #2554, opened 2018-04-27
EE-6039: Create Sockets connector release notes

ndinu:


connector-sockets.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+==== Min Mule Version
+*Mule 4.0.0*
+
+=== Fixed Issues
+
+[%header,cols="15a,85a"]
+|===
+|Issue |Description
+| https://www.mulesoft.org/jira/browse/MULE-14701[MULE-14701] | Connectors doesn't communicate correctly if their POJOs properties are nulleable or not
```
estebanwasinger: I would add a brief description explaining that now the metadata is more precise and communicates correctly when a property is nullable or not

estebanwasinger: (see which objects were affected, eg. attributes, etc)

PR #2569, opened 2018-05-03
MULE-15211 : Added inline example

ethanport:
WARNING Need to verify if this is true or I just made it up: `The correct type of this argument depends on what type is returned by the Java method.`.


dukesphere:
@estebanwasinger @ethanport **This has to be changed in the source comments for the Java connector**. Otherwise, it will get overwritten the next time we generated this file from that source.

PR #2626, opened 2018-05-21
Document post Crowd release approach

gramone:
Use post Crowd release approach to design and deploy APIs


george-hoffer:
Thank you Gilles!

PR #2627, opened 2018-05-22
We need to add Cloudhub Permissions

manibh:
We have added following two new permissions, we need to discuss with SME or CloudHub team and update the documentation accordingly with the definition of these permissions:
CloudHub Network Administrator
CloudHub Network Viewers


Mulesoft-Mysti:
@sleslie23 Okay to resolve conflict and merge?

PR #2632, opened 2018-05-22
more specific guidance on grouping process

sharkberto:
could be useful to have screenshot too - tbd 
Please edit if incorrect. In particular, @george-hoffer please sentence i wrote ....change "Group your APIs according to name and version only" to "Group your APIs by name only". I believe the group is created for APIs with the same name. Different versions of an API should have the same name.@antonkravchenko


george-hoffer:
Anton approved via private Slack conversation

PR #2635, opened 2018-05-23
Update java-invoke-method.adoc

PaulAndersonMule:


dukesphere:
@PaulAndersonMule Hi Paul, I get that the notes help identify the location where you see issues. However, this PR seems more like a JIRA issue than a PR, which is intended for merging new content into a file.

PaulAndersonMule:
I do it this way because filing a JIRA takes 5x longer to do, and also
slower for docs team to fix.

I'll stop PR in this style if you want, but I'm not filing JIRAs - don't
have the time for that

PR #2636, opened 2018-05-23
Added Hostname for Exchange assets

djuang1:
Not sure if this is going to change but Exchange assets are stored in S3 and the hostname needs to be whitelisted in order for policies to be applied.

### DIFF
--- a/runtime-manager/v/latest/installing-and-configuring-runtime-manager-agent.adoc
+++ b/runtime-manager/v/latest/installing-and-configuring-runtime-manager-agent.adoc
@@ -843,6 +843,7 @@ These tables show you the ports or IPs/hostnames to add to your whitelists to al
 |*mule-manager.anypoint.mulesoft.com* | 443
 |*analytics-ingest.anypoint.mulesoft.com* |  443
 |*arm-auth-proxy.prod.cloudhub.io* |  443
+|*exchange2-asset-manager-kprod.s3.amazonaws.com* |   443
 |===
 
 *Static IPs*


runtime-manager/v/latest/installing-and-configuring-runtime-manager-agent.adoc
```diff
@@ -843,6 +843,7 @@ These tables show you the ports or IPs/hostnames to add to your whitelists to al
 |*mule-manager.anypoint.mulesoft.com* | 443
 |*analytics-ingest.anypoint.mulesoft.com* |  443
 |*arm-auth-proxy.prod.cloudhub.io* |  443
+|*exchange2-asset-manager-kprod.s3.amazonaws.com* |   443
```
wpoch: That's an S3 bucket we shouldn't expose that endpoint directly.
@nicolasgnr @sramirezayuso @fpiagent could you review this, I think we are missing a platform DNS there.

djuang1: Thanks. Just some background. We ran a recent PoC at a prospect and we needed to whitelist that S3 bucket in order for API Policies to be applied correctly to an on-premise API.

Mulesoft-Mysti:
@wpoch Should I close this PR? Is some other change needed?

wpoch:
@Mulesoft-Mysti as I said in https://github.com/mulesoft/mulesoft-docs/pull/2636#discussion_r213408839 we need a platform DNS, and use that DNS in the docs. Given said that I should recommend you ping the people I mentioned, this PR is needed, but the values aren't right.

Mulesoft-Mysti:
@george-hoffer or @sleslie23  Can we resolve this? Anything I can do to help?

runtime-manager/v/latest/installing-and-configuring-runtime-manager-agent.adoc
nicolasgnr: So far the only way is to whitelist `exchange2-asset-manager-kprod.s3.amazonaws.com`. We are planning to add a custom domain + static IP.  What about `prod-eu`?

PR #2642, opened 2018-05-24
shared library doc error

PaulAndersonMule:


dukesphere:
created https://www.mulesoft.org/jira/browse/DOCS-2215 for this. PLG should be able to help with it once he is back from vacation.

PR #2672, opened 2018-06-01
preferred authentication valid values incorrect

jg-mule:
The preferred authentication valid values must be entered in uppercase, otherwise the application throws and error and will be unable to connect. Also raised a bug report on this as incorrect values are displaying in studio also


marianogonzalez:
That file is auto generated. Changes will be lost on next release. You should file a bug in the SDK's documentation generator

dukesphere:
@marianogonzalez Where is the JIRA project for the documentation generator?

PR #2690, opened 2018-06-07
EE-6112: Document how to use file global properties

mbuchwald:


mule4-user-guide/v/4.1/configuring-properties.adoc
```diff
@@ -127,6 +127,22 @@ To reference a variable that is defined in the OS, you can simply use the follow
 <logger message="${USER}" doc:name="Environment Property Set in OS" />
 ----
 
+=== File Properties
```
estebanwasinger: This is wrong placed.
1- This should not be inside `Environment Properties`
2- Should be placed just before `Environment Properties` with a higher hierarchy `== File Properties` instead of `=== File Properties`

mule4-user-guide/v/4.1/configuring-properties.adoc
```diff
@@ -127,6 +127,22 @@ To reference a variable that is defined in the OS, you can simply use the follow
...
+
+The placeholder value can come from a file. The placeholder value will became the string value of the mentioned file. For example:
```
estebanwasinger: `The placeholder value can come from a file`: so the value also can come from somewhere that is not a file?
I would rephrase it.

estebanwasinger: Would nice to explain why use File properties. E.g: extracting big contents, SQL Queries, transformations, etc.

mule4-user-guide/v/4.1/configuring-properties.adoc
```diff
@@ -127,6 +127,22 @@ To reference a variable that is defined in the OS, you can simply use the follow
...
+----
+Some content
+----
+
+[source,xml, linenums]
+----
+<mule:set-payload value="${file::properties-file.txt}"/>
+----
+
+Then the payload's value will be `"Some content"`. Just like properties files, these files must be located at `src/main/resources`, inside your Mule project, or you can also use absolute paths.
```
estebanwasinger: `Just like properties files` -> `Just like other properties files`

estebanwasinger: `inside your Mule project, or you can also use absolute paths.` -> `inside the Mule project, or also can be used absolute paths.`

PR #2699, opened 2018-06-08
DOCS-2231 app net visualizer

sleslie23:


sleslie23:
Received the following review comments from @dhilipuppili : 

Also i don’t see any thing on what happens when the application is re-deployed… (edited)
From what i understood the meta data for that application will be removed when the application structure is changed…
can we add that to the document as well please? This is important for the customers to know
Thanks..

PR #2702, opened 2018-06-11
Update salesforce-connector.adoc

tonedef71:
Not sure how SFTP configuration information is relevant to the Salesforce Connector.  If it truly is relevant, then the documentation should be updated to make such a relevancy more apparent to the reader; otherwise such SFTP documentation should be discarded from the Salesforce documentation altogether, as I have done here.


george-hoffer:
Thank you Anthony!!

PR #2716, opened 2018-06-14
Updated docs for Configuration Properties and Secure Configuration Properties

mbuchwald:


mule4-user-guide/v/4.1/secure-configuration-properties.adoc
```diff
@@ -121,6 +121,25 @@ Here is a flow that uses a non-encrypted value from the same secure properties f
...
+
+This way, it will depend on the value of the property `env` which file will be used to load the secure configuration properties. That `env` property could be set by a `global property`, or by a system property. 
+
+.Example: Secure Configuration Properties with environment file defined
+[source,xml, linenums]
+----
+<global-property name="env" value="dev"/>
+
+<secure-properties:config key="${runtime.property}" file="${env}-properties.yaml" name="test"/>
+----
```
pablolagreca: Just add some text saying that the default value will be taken from the global-property but later on can still be override with a system property or env variable

PR #2725, opened 2018-06-18
Update dataweave-cookbook-java-methods.adoc

adrianmartinmulesoft:
Added "One liner static call" because it's no explicitly mentioned and generated some question in https://mulesoft.slack.com/archives/C79H0F9HC/p1529096527000061

### DIFF
--- a/mule4-user-guide/v/4.1/dataweave-cookbook-java-methods.adoc
+++ b/mule4-user-guide/v/4.1/dataweave-cookbook-java-methods.adoc
@@ -52,6 +52,9 @@ This should output the following:
 [TIP]
 Instead of importing the class, you can import a single method from the class or all methods, just as when importing DataWeave modules.
 
+=== One liner static call
+It's possible to call a Java static method in a single line: `#[java!utils::MyUtils::appendRandom(vars.mystring)]`.
+
 == Instantiate a Java Class
 
 Through DataWeave code, you can instantiate a new object of any class. Note that after creating an instance, you can't call its instance methods via DataWeave, you can however reference its variables.


mule4-user-guide/v/4.1/dataweave-cookbook-java-methods.adoc
```diff
@@ -52,6 +52,9 @@ This should output the following:
 [TIP]
 Instead of importing the class, you can import a single method from the class or all methods, just as when importing DataWeave modules.
 
+=== One liner static call
```
Mulesoft-Mysti: It can be a weird reading experience to have just one subsection in a section, plus if your Java skills are rusty or weak (mine are both!), it takes a few reads to understand how much of the section above is "replaced" by importing and calling in one line. I recommend creating two subsections like this (I did it without headers, but you can do with if you think people need it):

You can call the method appendRandom from DataWeave code in three different ways:

1.  Import the class and refer to the method:

     <code sample>

2. Import one or more methods from the class, just as you do when importing DataWeave modules:

    <code sample same as #1 but with the import line changed>

3. Import and call the method in a single line: <code>

This option works for static methods only. 

The tip is really just another option, like importing and calling in the same line, so making it parallel helps the reader.

Mulesoft-Mysti: If this is more work than you planned, I'm happy to make the change myself (if you approve!)

Mulesoft-Mysti:
Closed, did this work plus a little more in https://github.com/mulesoft/mulesoft-docs/pull/2729. thanks for doing this!

PR #2740, opened 2018-06-20
Update anypoint-mq-connector-reference.adoc

manibh:
we should clarify the unit of acknowledgementTimeout which is milliseconds in this case


george-hoffer:
Fixed manually.

PR #2745, opened 2018-06-21
DOCS-2277 : Update Message Attributes documentation (Mule 4.1)

alegmarra:


alegmarra:
@george-hoffer Added the "code" element, thanks for the feedback.

PR #2767, opened 2018-06-25
DOCS-2286

Mulesoft-Mysti:
random asterisks removed.


george-hoffer:
This issue is not correctly addressed by this PR. The PR's fix only addresses a formatting issue and not the reason the text was entered incorrectly.

PR #2768, opened 2018-06-25
Typo5

Mulesoft-Mysti:
permission names are bolded in some places.


Mulesoft-Mysti:
Looks like someone is adding more content, so closing this.

PR #2788, opened 2018-06-28
DOCS-2266 additional changes to Anypoint VIsualizer for GA

sleslie23:


anypoint-visualizer/v/latest/technical.adoc
```diff
@@ -1,6 +1,6 @@
 = How Anypoint Visualizer Collects Application Data
 
-Anypoint Visualizer collects runtime information from running applications, APIs, and proxies to detect all of its dependencies. The following is a sample of the type of information that's being collected and sent to the control plane, where it is stored encrypted in a database.
+Anypoint Visualizer collects runtime information from running applications, APIs, and proxies to detect all of its dependencies. The following is a sample of the type of information that Anypoint Visualizer collects and sends to the control plane. This information stored encrypted in a database.
```
yaise: let's remove the event structure from the public documentation. This should remain private so we can be flexible about changing it.

sleslie23: @yaise I removed the "event" object and will reformat the remaining items as a table, not the raw JSON file.

PR #2819, opened 2018-07-03
Fix Graph Service CURL example to fit in one page

juancoen:
Also, added a longer query example split into multiple lines


george-hoffer:
Awesome!!  Thank you Juan.

PR #2829, opened 2018-07-05
Fix HTTP Caching Policy default values

balbifm:


api-manager/v/2.x/http-caching-policy.adoc
```diff
@@ -164,25 +164,25 @@ This parameter is optional and its default value is:
...
+#[attributes.method == 'GET' or attributes.method == 'HEAD']
 ----
 
-This means that only the responses for incoming requests with HTTP methods GET or CONNECT are cached.
+This means that only the responses for incoming requests with HTTP methods GET or HEAD are cached by default.
 
 === Conditional Response Caching Expression
 
-This field is a DataWeave expression to evaluate with the response, when evaluated, returns a boolean. If the expression returns true, the response tries to use the cache for potentially searching values in the cache, and storing the result. If the expression takes any other value, the cache won’t store the response. 
+This field is a DataWeave expression to evaluate with the response that when evaluated, returns a boolean. If the expression returns true, the response tries to use the cache for potentially searching values in the cache, and storing the result. If the expression takes any other value, the cache won’t store the response.
```
fermujica: I'd like to use a more straightforward description here. Would this be correct?

> This field holds a DataWeave expression that evaluates the response and returns a boolean. 
> If the response matches the expression, the expression returns True and the policy stores the response in the cache. 
> If the response takes any other value, the expression returns False and the cache won’t store the response. 

Let me know if I'm missing something.

balbifm: We should rephrase the other Conditional expression too, then

fermujica: Yes, I was thinking about that. If I'm not missing anything relevant, I'll simplify the request caching expression field and merge this.

api-manager/v/2.x/http-caching-policy.adoc
```diff
@@ -164,25 +164,25 @@ This parameter is optional and its default value is:
...
+This field is a DataWeave expression to evaluate with the response that when evaluated, returns a boolean. If the expression returns true, the response tries to use the cache for potentially searching values in the cache, and storing the result. If the expression takes any other value, the cache won’t store the response. 
 
 For information on DataWeave variables, see link:/mule4-user-guide/v/4.1/dataweave-variables-context[DataWeave Variables for Mule Runtime] and link:/mule4-user-guide/v/4.1/about-mule-message[Mule Message Structure].
 
 This parameter is optional and its default value is:
 
 [source,xml]
 ----
-#[attributes.statusCode < 300]
+#[[200, 203, 204, 206, 300, 301, 404, 405, 410, 414, 501] contains attributes.statusCode]
```
chiragnarangmulesoft: @balbifm @fermujica This change is not reflected on the docs site

fermujica: I just pushed this. 
This information will be updated within the next hour in our next docs site build.

PR #2832, opened 2018-07-05
DOCS-2493 Change to 4.x mule-artifact.json

ryan-slack-big:
There does not appear to be a place to move the 3.x particulars of this doc.


sleslie23:
@ryan-slack-big Thanks for the PR. We will triage this change and address it as soon as possible.

PR #2860, opened 2018-07-10
DOCS-2459 Adding CODEOWNERS file with initial content for testing.

sleslie23:


sleslie23:
@fermujica This is the file that maps owner to a specific directory. Any PR submitted to a directory should automatically be assigned.

.github/CODEOWNERS
```diff
@@ -0,0 +1,11 @@
...
+
+access-management/*    @sleslie23
+anypoint-b2b/*	@george-hoffer
+#anypoint-connector-devkit
+#anypoint-data-gateway
+#anypoint-exchange
+#anypoint-mq
+anypoint-platform-pcf/*		@sleslie23
+anypoint-private-cloud/*	@sleslie23
+anypoint-runtime-fabric/*	@sleslie23
```
vromero: I believe its fine for GH but usually its always a good practice to have a carriage return at the last line just in case the tooling uses GNU gettext

sleslie23: @vromero Thanks!

PR #2899, opened 2018-07-14
Fix a broken link to an image

rddheath:


rddheath:
I tested this change just now on a local build. However, I tested also before posting my original pull request for this content. Strange that the image didn't appear.

PR #2900, opened 2018-07-14
DOCS-2538 Update insight.adoc

veigap:


sleslie23:
@fermujica Would like your review on this also.

fermujica:
It makes sense. Looks good to me.

PR #2905, opened 2018-07-16
Add links to Studio download site from the installation instructions

fermujica:


anypoint-studio/v/5/download-and-launch-anypoint-studio.adoc
```diff
@@ -9,7 +9,7 @@
 
 To install Anypoint Studio:
 
-. Download Anypoint Studio from the link:https://www.mulesoft.com/platform/studio[Anypoint Studio Download Site] and extract it.
+. Download Anypoint Studio from the link:https://www.mulesoft.com/lp/dl/studio/previous[Anypoint Studio Download Site] and extract it.
```
sleslie23: @fermujica I noticed this is in v/5. I assume that is for Studio version 5. Does that link allow the user to download Studio 5?

anypoint-studio/v/6/download-and-launch-anypoint-studio.adoc
```diff
@@ -4,7 +4,7 @@ Before downloading and launching Anypoint Studio, make sure you have the Java SE
 
 To install Anypoint Studio:
 
-. Download the Anypoint Studio zip file from the download site and extract it
+. Download Anypoint Studio from the link:https://www.mulesoft.com/lp/dl/studio/previous[Anypoint Studio Download Site] and extract it.
```
sleslie23: @fermujica I only see 6.5 from the dropdown.

anypoint-studio/v/5/download-and-launch-anypoint-studio.adoc
fermujica: No. I don't think there is a download site for Studio 5 anymore. We'll archive this site in a few more days, so I added this link temporarily in case someone lands here from outside searches.

anypoint-studio/v/6/download-and-launch-anypoint-studio.adoc
fermujica: Same applies here. I don't think we have a download site for Studio 6, and we'll archive this article in the future.

anypoint-studio/v/5/download-and-launch-anypoint-studio.adoc
fermujica: I see. OK, good point. We don't have an open download site for these versions. 
I'll rewrite the topics for these versions to reflect that and suggest to download the newer versions instead.

fermujica: I removed download instructions for Studio 5. See 605d6f6bf

anypoint-studio/v/6/download-and-launch-anypoint-studio.adoc
fermujica: I'm pointing users from Studio 6.4 and below to the download and installation instructions for 6.5. See 0fd1a318a

PR #2911, opened 2018-07-16
DOCS-2563 Update cloudhub-runtimes-release-notes.adoc

fededonna:


sundippatel:
Can we be more descriptive on what's being changed? What bugs did we fix? What changes does the Object Store v2 update bring?

For example, in a previous update (April 5th), we wrote the following:

* Updated the Object Store V2 plugin to provide better error messages

PR #2831, opened 2018-07-05
Edge Tasks for Anypoint Runtime Fabric.

fermujica:

PR #2920, opened 2018-07-18
DOCS-2207 RTF initial doc review

sleslie23:
@mcocaro @sundippatel Please add any additional reviewers as required.


anypoint-runtime-fabric/v/1.0/configure-edge.adoc
```diff
@@ -0,0 +1,40 @@
+= Enable Edge for Anypoint Runtime Fabric
```
fermujica: This article should use the latest and reviewed version: https://github.com/mulesoft/mulesoft-docs/pull/2831

PR #2922, opened 2018-07-18
DOCS-2584 Update roles.adoc

fermujica:
Adding information about specific behavior by design which is not documented: https://www.mulesoft.org/jira/browse/SE-8707

### DIFF
--- a/access-management/v/latest/roles.adoc
+++ b/access-management/v/latest/roles.adoc
@@ -17,7 +17,7 @@ image::roles-a3471.png[roles-a3471]
 Please be aware that roles are assigned per Business Group (Business Group specific).
 
 [NOTE]
-When a new Business Group is created, all the Organization Administrators of the parent business group are also added as Organization Administrators in the newly created Business Group (child). However, when creating a new user as an Organization Administrator in the parent Business Group, this role is not propagated to all the sub Business Groups.
+When a new business group is created, all the users with the Organization Administrators role in the parent business group are also added as organization administrators in the newly created business group (the child business group). However, when creating a new user as an organization administrator in the parent business group, this role is not propagated to all the sub business groups.
 
 == Default Roles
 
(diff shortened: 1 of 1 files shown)


access-management/v/latest/roles.adoc
```diff
@@ -14,6 +14,11 @@ To access the Roles menu in Access Management, ensure that you are in the correc
 
 image::roles-a3471.png[roles-a3471]
 
+Please be aware that Roles are assigned per Business Group (Business Group specific).
+
+[NOTE]
+When a new Business Group is created, all the Organization Administrators of the parent business groups are also added as Organization Administrators in the newly created Business Group (child). However, when creating a new user as an Organization Administrator in the parent Business Group, this role is not propagated to all the sub Business Groups.
```
sleslie23: I don't think we should capitalize Organization Administrators since it is not a product or UI element. Not sure about business group.

access-management/v/latest/roles.adoc
```diff
@@ -14,6 +14,11 @@ To access the Roles menu in Access Management, ensure that you are in the correc
 
 image::roles-a3471.png[roles-a3471]
 
+Please be aware that Roles are assigned per Business Group (Business Group specific).
```
sleslie23: Roles should not be capitalized unless it is a UI element.

access-management/v/latest/roles.adoc
arielmira: The first sentence of the documentation states: "This topic describes how use roles to manage permissions within Anypoint Platform. This topic assumes that you have an Organization Administrator role in your organization or that you have API Versions Owner permissions and want to manage user permissions for an API version." And Organization Administrator is with capital letters there to emphasize the role name in the platform. I applied the same logic to Business Groups (to remark they official name of them). However, feel free to use only minus if needed.

PR #2944, opened 2018-07-20
DOCS-2541 wording fix

george-hoffer:


george-hoffer:
This change was superceded by a better wording change and is no longer needed.

PR #2945, opened 2018-07-20
Deployment with Runtime Fabric

sbehrends:


fermujica:
I'm moving this to #2985

PR #2946, opened 2018-07-20
Change create attribute to false

PabloCamino:
This documentation is for API Manager 2.0, so it can't have a create="true" attribute


Mulesoft-Mysti:
It looks like the example changed, and moved to /mule4-user-guide/v/4.1/migration-example-complex.adoc. @sleslie23 or @rddheath , is this change still needed?

PR #2963, opened 2018-07-24
Docs 2535 apid 2.2.8

rddheath:
This pull request contains the release notes for API Designer 2.2.8, which is the newer version of APID. There was a file for the release notes for the older version of APID that is built into API Manager, but no file for the release notes for the newer version that is in Design Center. In fact, the Design Center release notes really were for flow designer.

So, I split the Design Center release notes into three files:
1. A file for the release notes for APID.
2. A file for the release notes for flow designer.
3. A parent file for the Design Center that links to both of the previous files.

I also added release notes for several previous releases of APID because those sets of release notes had not yet been written.


iamerico:
**1) Error messages were not appearing for invalid examples nodes.**
We could add the use case that triggered this bug.
In the following definition, the Examples node is wrong
```#%RAML 1.0
title: New API

types:
  ids:
    type: object
    properties:
      identification:
        type: string
        examples:
          - MuleSfot
          - Salesforce
```

The correct way to define Examples is:
```#%RAML 1.0
title: New API

types:
  ids:
    type: object
    properties:
      identification:
        type: string
        examples:
          id1: MuleSfot
          id2: Salesforce
```

The - symbol is used to define array elements.
```#%RAML 1.0
title: New API

types:
  ids:
    type: object
    properties:
      identification:
        type: array
        examples:
          id1:
            - MuleSoft
            - Salesfroce
          id2:
            - Mule
            - It comes out
```

**2) When switched on, the mocking service could incorrectly insert the URI to the mocking server.**
It could be added. We are now supporting the declaration of the baseUri with the following structure:
```
baseUri:
  value: http://www.example.com/api
```
**_This is valid only for RAML 1.0, not for RAML 0.8_**
This is valid according to the specification https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#annotating-scalar-valued-nodes

rddheath:
About 2: I wasn’t freewheeling, but doing what Kris had done in the Design Center release notes that she wrote. I’ll fix the headings, though.

About 7: 1. Kris wrote the notes for all releases before April 14 this year. That’s when I joined MuleSoft. 2. My text editor finds no instances of “auto complete” or “auto completion”.

About 8: Kris wrote this section, which contains release notes for API Designer. It was written back in 2017, which is why I decided not to revise it. I doubt anyone will look back this far in the topic to see release notes from August 2017.

About 9: Again, Kris wrote this in 2017, so it is outside of the scope of my changes so far in 2018. Nobody will look for release notes that old, anyway.

About 10: this hasn’t caused any “blowouts” since version 1 was released on July 20, 2017, apparently. Also, I did just run a local build to check and it shows up without a problem.

PR #2968, opened 2018-07-25
Docs 2374

dukesphere:
Updated conceptual info on Flow and Subflow (as part of DOCS-2374 and DOCS-2375)


mule4-user-guide/v/4.1/about-flows.adoc
```diff
@@ -1,26 +1,74 @@
...
+flows and subflows that you add to the app and link together, often with
+link:flowref-about[Flow Ref] components. Mule apps in production environments
+typically use multiple flows and subflows to divide the app into functional
+modules or for <<error_handling, error-handling>> purposes. For example, one
+flow might receive a record and transform data into a given format that another
+flow processes in some special way.
 
-image::about-flows-ef7ca.png[about-flows-ef7ca]
+In addition to calling flows or subflows through the Flow Ref, you can
+also use a Connector event processor in a flow or a DataWeave `lookup` function,
```
afelisatti: AFAIK there are no components but flow ref and lookup that allow executing a flow (apikit is a very special case not worth documenting).

mule4-user-guide/v/4.1/about-flows.adoc
```diff
@@ -1,26 +1,74 @@
...
+appropriate response. If the event processing gets complicated, or must call
+out to other services, you might factor out that behavior into other (perhaps
+reusable) flows.
 
-image:about-flows-98a17.png[about-flows-98a17]
+[[error_handling]]
+Each flow (but not subflow) can have its own error handling. One reason to
+call a flow through a Flow Ref component is to separate out the error
+handling to different contexts. For example, a parent flow that receives web
+client requests might define HTTP-related error handling. If the parent flow
```
afelisatti: Not sure about this, it's a contrast to the existence of the try component which in this example would allow you to deal with everything JMS related right there.

mule4-user-guide/v/4.1/about-flows.adoc
```diff
@@ -1,26 +1,74 @@
...
 
-Note that flows always function synchronously. If you need to achieve asynchronous patterns, such as Fire-and-Forget, you can use the Async (`<async>`) component.
+Like a flow or subflow, a link:try-scope-concept[Try scope] also groups
+together a sequence of event processors with its own error handling. You can
+place a Try scope inside a flow or subflow to isolate error handling inside a
+flow without the need to create a separate, private flow or subflow. The trade
+off is that error handling through a Try scope occurs inline inside the flow,
+which makes it harder to reuse between other flows or subflows.
+
+In both cases, calling a private flow or including a Try scope isolates the
```
afelisatti: what's the meaning of the private part? AFAIK it's not a concept we handle

dukesphere: Oh, training uses the term "private flow" for a flow that does not have an event source.

mule4-user-guide/v/4.1/about-flows.adoc
```diff
@@ -1,26 +1,74 @@
...
+flows and subflows that you add to the app and link together, often with
+link:flowref-about[Flow Ref] components. Mule apps in production environments
+typically use multiple flows and subflows to divide the app into functional
+modules or for <<error_handling, error-handling>> purposes. For example, one
+flow might receive a record and transform data into a given format that another
+flow processes in some special way.
 
-image::about-flows-ef7ca.png[about-flows-ef7ca]
+In addition to calling flows or subflows through the Flow Ref, you can
+also use DataWeave `lookup` function in some component fields to pass the
```
afelisatti: instead of `in some components` I would just say `within expressions and transformations` which is more accurate and still covers the fact that a component would need to support those to make use of it.

mule4-user-guide/v/4.1/about-flows.adoc
dukesphere: I removed "private flow".

mule4-user-guide/v/4.1/about-flows.adoc
```diff
@@ -1,26 +1,86 @@
...
+flow or within a Try scope results in exactly the same error handling 
+behavior, but the event processors in a Try scope cannot be called from 
+another flow.
+
+== Branching Event Processing
+
+Some scopes are available for a flow to branch event processing into separate
+threads and to allow asynchronous processing of the event at a specific point
+in the flow, for example, the link:scatter-gather-concept[Scatter-Gather] and
+link:async-scope-reference[Async] scopes. A link:choice-router-concept[Choice]
```
afelisatti: Here I would say that processing can also be branched according to some condition as with the choice component.

mule4-user-guide/v/4.1/about-flows.adoc
```diff
@@ -1,26 +1,86 @@
...
+You typically use Flow Ref component or `lookup` function only for 
+triggering flows that lack a source. 
 
-There are two ways of connecting flows:
+Like functions or methods in other programming languages, it is a best practice
+to focus your flows on specific (perhaps reusable) activities, such as receiving
+an API request from a web client, processing the event, then returning an
+appropriate response. If the event processing gets complicated, or must call
+out to other services, you might factor out that behavior into other (perhaps
+reusable) flows.
```
afelisatti: Shouldn't this be rephrased to avoid the same "perhaps reusable" thing?

PR #2971, opened 2018-07-26
Exchange doc fix per Slack user note

george-hoffer:


george-hoffer:
Too many issues in this addition per review. No longer worth pursuing.

PR #2975, opened 2018-07-26
MQ-1202 note

george-hoffer:


anypoint-mq/v/latest/index.adoc
```diff
@@ -16,7 +16,8 @@ link:/anypoint-mq/mq-apis[REST APIs],
 link:/anypoint-mq/mq-faq[FAQ], and 
 link:/release-notes/anypoint-mq-release-notes[Release Notes].
 
-*Important:* Do not use Anypoint MQ connector version 1.1.1. Customers should upgrade to MQ connector version 1.1.2 or later as soon as possible.
+*Note:* Anypoint MQ does not support use with the Insight feature of Runtime Manager. 
+(Accessed from Anypoint Platform > Runtime Manager > application > Manage Application > Insight.)
```
sleslie23: I don't think you should document the path to the Insight tab here. There are procedures that mention that in the Runtime Manager docs. Also, I would say that insight is a feature of CloudHub, not Runtime Manager.

PR #2991, opened 2018-07-30
DOCS-2206 Update install-aws.adoc

jeff-1amstudios:


sleslie23:
Thanks @jeff-1amstudios

PR #3010, opened 2018-08-01
Clarify that maven install is required

adobni:
m2 replaced by Maven repository.
Clarify that building and testing in the same execution doesn't work.


Mulesoft-Mysti:
@fermujica I can fix the "an" to be "a". Is this okay to merge with that change?

PR #3018, opened 2018-08-01
updated landing page--stage 1

Mulesoft-Mysti:


Mulesoft-Mysti:
Note that the goal isn't to replicate the ToC, it's to give people a suggestion as to a great place to start.

index.adoc
```diff
@@ -1,84 +1,53 @@
...
-* link:https://anypoint.mulesoft.com/login/#/signup?apintent=exchange[Create an Anypoint Platform account].
-* Consider the link:/mule-user-guide/v/3.8/hardware-and-software-requirements[MuleSoft hardware and software requirements]
-* For development of Mule apps or API design, set up your workstation according to the link:#getting-started-dev[Developer Toolkit].
-* To manage and deploy Mule apps, see the link:#getting-started-deploy[Deployment Toolkit].
-* To set up platform permissions or monitor deployed assets see the link:#getting-started-manage[Administrator Toolkit].
+You can deploy locally or in the cloud, using a variety of methods.
 
+|===
+| **Deploying**     | **Testing** | **Managing**
+| link:runtime-manager/deployment-strategies[Choosing a deployment strategy] (will point to Steven's tutorial)
```
Mulesoft-Mysti: Mysti to remove paranthetical comment (just for reviewers info)

index.adoc
```diff
@@ -1,84 +1,53 @@
...
 
-[[what-is]]
-== What is Anypoint Platform?
+== Designing and Building Integration Apps and APIs
 
-Anypoint Platform is MuleSoft's *integration application and API suite*. It comprises a unique toolset that enables organizations to create, integrate, model, build, and deploy services, APIs, and Mule applications.
+Anypoint Studio is the Eclipse-based IDE to help you create apps and APIs, and Design Center is the cloud-based tool.
+|===
+| **Building Apps** | **Building APIs**
+| link:/anypoint-studio/v/7.1[Building Apps in Anypoint Studio] (will point to quickstart soon) | link:/anypoint-studio/v/7.1/import-api-def-dc[Importing APIs from Design Center to Anypoint Studio]
```
Mulesoft-Mysti: Mysti to remove parenthetical (just for reviewers info).

dukesphere:
It looks a heck of a lot better and is an improvement. I am not convinced that these are the most natural categories/divisions for our doc. (1) App development is strongly tied to the topics listed in the Mule Runtime 3 & 4 section. Specifically, connectors, components, DW, etc. are essential to Mule app creation. (2) On the other hand, I think of API design as its own category, separate from Mule app creation. API design can be a precursor to app development but need not be. You can create an app without designing an API, and the audience (or at least the kinds of tasks you need to be able to perform) for API design are different than for the Mule app. (3) Most of the Runtime topics are really refs for app dev rather than deployment, though the intro seems to be emphasizing deployment.  (4) Exchange is a See Also, but the sharing use case is essential to our product offering.

dukesphere:
I didn't mean to send that yet. My last thought is that I'm happy to hear the rationale behind that org and am open to other thoughts around it, of course.

rddheath:
1. If we're going to place a marketing slogan after the title, the slogan should be larger than the text used in the body, and perhaps be in bold. As it is, it looks like regular content on the page.
2. Isn't Anypoint Platform only the MuleSoft products that are online? Anypoint Studio is not part of the platform, but is a standalone IDE.
3. "Use the Anypoint Platform to build an application network with secure, reusable integrations and APIs." : It seems awkward to use "integrations" by itself without saying what is being integrated.
4. "Designing and Building Integration Apps and APIs": what is the difference here between "designing" and "building"? "Building" in the dev world usually means something similar to "compiling". And, though the heading mentions "designing", the rest of this section doesn't refer to designing, but only to building. That seems inconsistent. "Building", in fact, seems inconsistent with the terminology used in Design Center, which is "designing".
5. "Mule Runtime 3 and Mule Runtime 4": This heading doesn't use gerunds, as the other headings do. What do you do with the runtime? Can the heading explain that with a gerund or two?
6. "You can deploy your integration apps and APIs locally, using the standalone Mule Runtime (Mule).": So, you deploy with runtime? Why not use the gerund "deploying" in the heading? It seems odd, too, that this sentence says that you use runtime for deploying, yet the next section starts with a heading that begins with "deploying".
Also, I would change "(Mule)" to (aka Mule)".
7. "The connectors and DataWeave version you need depend on the version of Mule you use.": Need to do what? Deploy? The previous sentence is about deploying, so "you need" naturally uses that as its antecedent.
Does the first instance of "version" refer back both to "connectors" and to "DataWeave"? So, you mean something like this: "The version of each connector that you use to build an app and the version of DataWeave that you use to transform messages in your app depend on the version of Mule that you use."?
8. It seems weird to have the first field in the Mule 3 column of the table blank. Usually, if there are more rows with content in one column than in another column, the content of the column with fewer rows is shifted up, not down. If you really want to keep the content in the two columns parallel, how about moving "Mule 4 for Mule 3 users" to a sentence before or after the table?
9. There is a spelling error in "Deploying, Managaing, and Testing Your Integrations".
10. "You can deploy locally or in the cloud, using a variety of methods.": The heading mentions deploying, managing, and testing. Why doesn't the paragraph after the heading mention all three, too? Right now, it mentions only deploying.
11. Are those three links in the Testing column? If they are, why not put them on separate lines?
12. It seems that the content in the Testing and Managing columns should be switched. 
13. "In addition to the quickstarts in the developer site, we provide documentation for developer kits and platform tools.": In previous sections, the paragraphs following the headings elaborated upon those headings a bit. This paragraph doesn't elaborate on its heading, so I'm left wondering what I can develop for the Anypoint Platform. Shouldn't the paragraph explain that before stating that there is documentation to help you in the task of developing?
14. In the columns Developer Kits and Platform Tools, are those three links joined together? That looks confusing. How about separating the links on to separate rows?
15. "More Resources/Anypoint Exchange": Why is this the only mention of one of the central tools of the platform? Exchange allows people to share the APIs and apps that they create, which is the whole purpose of creating them.

rddheath:
16. "When an older version of a product is no longer supported, incuding those with end-of-life status, we move the documentation to https://docs.archive.mulesoft.com/.": 
a. There is a spelling error: "incuding".
b. What does "those" refer to? If its antecedent is "older version", then they conflict because "those" is plural and its antecedent is singular. You could rewrite the sentence like this: "When older versions of a product are no longer supported because they have reached their end of life, we move their documentation to ...".

rddheath:
17. "Your positive impact in the community by improving MuleSoft documentation is appreciated and recognized.": 
a. "impact by"? I don't think "impact" takes that preposition.
b. "recognized": Such a use of "recognized" usually is in the context of giving somebody something in return for their doing something. For example, "In recognition of your years of service, we give you this gold watch." I think the sentence should end at "appreciated".

dukesphere:
We also have two important Monitoring products:
* Anypoint Monitoring
* API Functional Monitoring

Mulesoft-Mysti:
Here's to top part of the revised version: 
image

PR #3037, opened 2018-08-03
Document updated mule-artifact.json management in Studio 7.2

fermujica:
This article is meant to help users understand the workspace validation that Studio 7.2 runs to update certain declarations in their projects to use the latest Mule Maven Plugin.


anypoint-studio/v/7.2/update-workspace.adoc
```diff
@@ -0,0 +1,69 @@
...
+|
+[source,json,linenums]
+----
+{
+  "minMuleVersion": "4.0.0",
+}
+----
+|===
+
+Note that this simplification is meant to reduce the amount of declarations you need to keep track at development time using Anypoint Studio. A fully configured mule-artifact.json is still required by the runtime when executing the packaged application. You can find the generated mule-artifact.json file inside the `target` directory after packaging.
```
Mulesoft-Mysti: "number" not "amount" (amount is for mass nouns like 'rice' or 'hay').

anypoint-studio/v/7.2/update-workspace.adoc
```diff
@@ -0,0 +1,69 @@
+= Updating Workspaces
+
+Anypoint Studio 7.2 simplifies the configuration of the `mule-artifact.json` file. This file is used when packaging your Mule application to identify the class loader configuration and settings of your Mule application, and the minimum required Mule runtime version to run your application. +
```
Mulesoft-Mysti: We should add tick marks to every instance of `mule-artifact.json`, not just this one.

anypoint-studio/v/7.2/update-workspace.adoc
```diff
@@ -0,0 +1,69 @@
...
+{
+  "minMuleVersion": "4.0.0",
+}
+----
+|===
+
+Note that this simplification is meant to reduce the amount of declarations you need to keep track at development time using Anypoint Studio. A fully configured mule-artifact.json is still required by the runtime when executing the packaged application. You can find the generated mule-artifact.json file inside the `target` directory after packaging.
+
+The Mule Maven plugin checks the metadata declared in mule-artifact.json. If the only declared value is `minMuleVersion`, the Mule Maven Plugin infers all the other metadata from the application code before packaging.
+If there are other metadata declared, it uses that information to package the application. +
```
Mulesoft-Mysti: "If other metadata is declared" (less awkward), "the plugin" instead of "it" because the antecedent isn't clear.

anypoint-studio/v/7.2/update-workspace.adoc
```diff
@@ -0,0 +1,69 @@
...
+If you choose to update to the latest Mule Maven plugin, and to keep your old descriptor files, you'll need to manually keep track of all your descriptors metadata.
+
+== Upgrading From 7.1.x to 7.2
+
+When you open Anypoint Studio 7.2 and use a workspace created with previous versions, Studio pops up a note:
+
+image::this-766bb.png[align=center]
+
+Note that Studio lists all the projects whose configurations needs to update. +
```
Mulesoft-Mysti: "whose configurations need updating" or "whose configurations will be updated" as 'needs to update' to avoid the missing subject problem (whose configurations Studio needs to update).

anypoint-studio/v/7.2/update-workspace.adoc
Mulesoft-Mysti: Adding "there" doesn't fix the problem of metadata being a mass noun and the verb being plural. Also, "it" has no antecedent so it sounds like the metadata is using its own info to package the app. I believe the plugin is the "it".

Mulesoft-Mysti: Can we fix line 47?

PR #3045, opened 2018-08-05
DOCS-1981 VPN Doc Review

sleslie23:


runtime-manager/v/latest/vpn-about.adoc
```diff
@@ -0,0 +1,37 @@
+= About CloudHub VPN
```
sebastiankorol: Anypoint VPN

runtime-manager/v/latest/vpn-about.adoc
```diff
@@ -0,0 +1,37 @@
...
+The number of VPNs you can create depends on the VPN capacity allocated to your account.
+
+The to create and maintain a secure connection with your external network, the CloudHub VPN provides the following features:
+
+* Route-based access via 2-byte and 4-byte Autonomous System Numbers (ASN). This defines the routing policies used to connect to your network through the VPN.
+* Reusable IP addresses
+* Tunnel configuration 
+
+
+*Note*: Network Address Translator (NAT) is not supported.
```
sebastiankorol: NAT stands for Network Address Translation, and we'd need to explain what it means in this context. What you're trying to say is that the VPN gateway doesn't do NAT. What most people would understand in the context of VPN is that it doesn't support NAT traversal, which it does.

runtime-manager/v/latest/vpn-create.adoc
```diff
@@ -0,0 +1,70 @@
+= Create a CloudHub VPN 
+
+This topic describes how to create a CloudHub VPN using Runtime Manager. For information on creating a VPN using the CloudHub API see link:https://anypoint.mulesoft.com/exchange/portals/anypoint-platform-eng/f1e97bc6-315a-4490-82a7-23abe036327a.anypoint-platform/cloudhub-api/[CloudHub API].
+
+== Prerequisites
+
+Before creating a VPN, you must create a Virtual Private Cloud (VPC). See link:/runtime-manager/virtual-private-cloud[About Virtiual Private Cloud] for more information. You should also ensure that the components of your external network are configured correctly.
+
+Additionally, you must provide the following when creating or configuring a VPN:
```
sebastiankorol: PSKs, PTP, and local ASNs are optional

runtime-manager/v/latest/vpn-create.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+* *Remote IP address*: the IP address of the device you are connecting to.
+* *Pre-shared Key (PSK)*: the shared secret used by the network tunnel that connects between Anypoint Platform and your network.
+* *Point-to-Point CIDR*: the IP range used by the network tunnel. You must configure a unique IP address range for each tunnel.
+* *Remote and Local Autonomous System Number (ASN)*: specifies the collection of routing prefixes. You must configure both a remote and local ASN before creating a VPN. When creating a VPN, you must provide the IP address of both.
+* *Classless Inter-Domain Routing (CIDR)*: specifies the routing address of your network. This is required only if you are using static routing.
```
sebastiankorol: This is not a field in the form, is it? It should be `static routes`, which is a list of subnets to route over the VPN connection, not a single address

runtime-manager/v/latest/vpn-create.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+. From the Mangagment Center, select Runtime Manager.
+. Go to the environment where you want to create a VPN.
+. Click the *VPN* tab, then click *Create VPN*.
+. Enter a name for your VPN.
+. Select a Virtual Private Cloud from the drop down list.
+. Enter the remote IP address used to connect to your VPN.
+. Select the routing type you want to use for your VPN.
++
+We recommend that you use Border Gateway Protocol (BGP) routing when possible. See link:/runtime-manager/vpn-routing[About CloudHub VPN Routing] for more information.
```
sebastiankorol: Can we explain why?

runtime-manager/v/latest/vpn-create.adoc
```diff
@@ -0,0 +1,70 @@
...
+To use BGP routing:
++
+ .. Select *BGP Routing*.
+ .. Enter the port of your local ASN. The default value is 65000.
++
+To use static routing:
++ 
+ .. Select *Static Routing*.
+ .. Click *Add New Rule*.
+ .. Enter a Classless Inter-Domain Routing (CIDR) address. For example:
```
sebastiankorol: Network, not address. It should read: _Enter a network in CIDR notation_

runtime-manager/v/latest/vpn-create.adoc
```diff
@@ -0,0 +1,70 @@
...
+You must configure two tunnels that connect between different endpoints. You can use the same PSK value for both connections, but you must use different IP ranges.
++
+.. Select the type of tunnel configuration for your VPN:
++
+** *Automatic*: automatically configures the required tunnels for your VPN using the default values.
+** *Custom*: enables you to manually configure the tunnels used by your VPN.
++
+See link:/runtime-manager/vpn-tunnels[About CloudHub VPN Tunnels] for more information.
++
+.. Enter a vallue for the  PSK.
```
sebastiankorol: nit: value and extra space before PSK

runtime-manager/v/latest/vpn-create.adoc
```diff
@@ -0,0 +1,70 @@
...
+ .. Click *Add New Rule*.
+ .. Enter a Classless Inter-Domain Routing (CIDR) address. For example:
+ +
+ ----
+ 10.0.0.0/8
+ ----
+
+. Configure the tunnel used to connect between Anypoint Platform and your network.
++
+You must configure two tunnels that connect between different endpoints. You can use the same PSK value for both connections, but you must use different IP ranges.
```
sebastiankorol: We configure endpoints, not tunnels. Each endpoint already has two tunnels, so configuring a single one is enough, and IP ranges (for static routes, I assume?) aren't different -- they're a property of the endpoint, not the tunnel

sleslie23: @sebastiankorol Are you sure the UI allows you to only configure one? My understanding from Vinay and Daniel is that it is not.

sebastiankorol: You configure endpoints, not tunnels. The doc should reflect that. Also, it's not clear what "different IP ranges" is referring to. Static routes? Then no, it's a property of the endpoint. PTP addresses? Then "you must configure..." is inaccurate since that's optional, and for 99% of the use cases, customers shouldn't be filling them in at all.

runtime-manager/v/latest/vpn-about.adoc
sleslie23: Updated globally

sleslie23:
Here are Henry's review comments:

Hello Steven,

I looked over the documentation and I have some feedback:

"If required, you can create multiple VPNs to connect to different networks or different regions within a network." This needs to be clarified as distinct from VPC peering. There are no AWS VPCs that can be spread across regions. If this means connecting two Cloudhub VPCs, then the user must configure their own network and connect to both VPCs, configuring routes either statically or propagating them via BGP.

"Route-based access via 2-byte and 4-byte Autonomous System Numbers (ASN). This defines the routing policies used to connect to your network through the VPN."

This does not define the routing policies. I suggest saying "the user's network must be identified by a 2-byte ASN, while the ASN for your Cloudhub VPC supports 2-byte ASNs and 4-byte ASNs in the 4200000000 to 4294967294 range."

Reusable IP addresses - I don't think anyone will understand this without context - I propose, 'multiple gateway IP addresses from your network can be used.'

Additionally, there should be some documentation that states that once the first VPN Connection for your Cloudhub VPC is created, you can only use the Local ASN entered for every subsequent VPN Connection. I.E., that the ASN will identify your Cloudhub VPC network to every other HA VPN Connection.

NAT is Network Address Translation, changing from Network Address Translator

For feature Tunnel Configuration, there needs to be some detail. The users will be able to configure their tunnel shared secret and point-to-point CIDR (must be in the 169.254.0.0/30 range and must not include the networks identified here: https://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_VPN.html#VPNTunnels)

Typo in "The to create and maintain"

Supported Data Encryption Methods

Typo in "bewtween"

Please add that the SHA-2 hashing algorithm verifies network traffic integrity

Please specify the specific DH groups that the Cloudhub VPN supports for the key exchange (same as AWS)

Also, to use these features, please specify that the user only has to configure their network appliances with these settings, no other configuration is necessary.

"This allows the device you are connecting to broadcast routes to the VPN." - Replace with 'advertise' routes

Thanks!

PR #3059, opened 2018-08-07
Docs 2523 migrator

dukesphere:
very early draft, more content coming. This is for a preliminary look by Rodro and hope to get some early feedback.  

Possible TODOs (already Slacked to Rodro): I am thinking of adding a list of components, etc. the migrator will support when released.  Please let me know if you think this is going in the right direction and any changes you would advise. Also, I would like to be able to point people to the Compatibility module so they can use it as a temporary “bridge” before the full migration.


mule4-user-guide/v/4.1/migration-tool.adoc
```diff
@@ -0,0 +1,216 @@
+= Migration Tool
+
+The Migration tool (migrator) is a command-line utility that helps facilitate migrations of Mule 3 projects (Mule apps written for the Mule 3 runtime) to
```
elrodro83: Even if this is a command line, only for the BETA will the user actually execute the command. The idea is that for GA it is invoked from Studio.

Is this document intended to be just for Beta or or to apply to GA also?

mule4-user-guide/v/4.1/migration-tool.adoc
```diff
@@ -0,0 +1,216 @@
...
+Mule 4. Instead of performing a completely manual migration, you can use the
+tool to handle the migration components, connector, and module configurations
```
elrodro83: `to handle the migration ` -> `to perform a first automatic step for the migration of `

It should be very clear that running the tool is just the first step and no every scenario can be handled by the tool. There is a manual step(s) afterwards

mule4-user-guide/v/4.1/migration-tool.adoc
```diff
@@ -0,0 +1,216 @@
...
+from Mule 3 to Mule 4.
+
+The migrator runs a sequence of tasks over a Mule project and then outputs
+a migrated project with a report about the migration.
```
elrodro83: `migrated project` -> `Mule 4 project`

mule4-user-guide/v/4.1/migration-tool.adoc
```diff
@@ -0,0 +1,216 @@
...
+
+//TODO: FROM RODRO: we want them to migrate to a point where the compatibility plugin can be dropped from an application
+
+//TODO: QUESTION: WILL THEY ASK PEOPLE IF THEY WANT TO MIGRATE CASES WHERE BEST PRACTICES ARE NOT FOLLOWED?
+
+//TODO: QUESTION: A LOT ISN'T COVERED. ARE THE CONNECTOR TEAMS.
+//TODO: DO WE HAVE DOC ON THE COMPATIBILITY MODULE?
+
+
+Note that the Migration tool is different from the Compatibility module, which
```
elrodro83: The Compatibility plugin is a complement of the Migration tool.

It provides MEL support, for the cases where a MEL expression could not be migrated to DW automatically and also provided adaption between the different message models of Mule 3 and Mule 4.

mule4-user-guide/v/4.1/migration-tool.adoc
```diff
@@ -0,0 +1,216 @@
...
+is for supporting some key Mule 3 functionality, such as MEL expressions, in
+Mule 4 projects. The Compatibility module is meant as a temporary bridge for
+to use until you complete your migration, and it does not cover all Mule 4
+functionality.
+
+//TODO: QUESTION: API for the tool? Tool consists of an execution engine,
```
elrodro83: The end user wouldn't care about this.

mule4-user-guide/v/4.1/migration-tool.adoc
```diff
@@ -0,0 +1,216 @@
...
+//* Step: An operation that changes, removes, or updates a resource or content
+//in a Mule project.
+
+== Requirements
+
+[[min_reqs]]
+Minimum requirements:
+
+* Java 8
+* Maven 3.3.9
```
elrodro83: the migrator itself doesn't use maven

mule4-user-guide/v/4.1/migration-tool.adoc
```diff
@@ -0,0 +1,216 @@
...
+//TODO: PREREQUISITES
+
+To run the Migration tool:
+
+. Make sure that the required software is installed. See <<min_reqs>>.
+. In your console, provide a command that specifies all the required <<options>>, for example:
++
+----
+$ java -jar
+ mule-migration-tool-runner-1.0.0-20180806.163412-176-jar-with-dependencies.jar
```
elrodro83: the actual released jar will have a shorter name (the one we publish for download)

mule4-user-guide/v/4.1/migration-tool.adoc
```diff
@@ -0,0 +1,216 @@
...
+Executing migration...
+...
+===========================================================================
+MIGRATION SUCCESS
+===========================================================================
+Total time: 11.335 s
+Migration report:
+/Users/me/AnypointStudio/v7/my-migrated-project/report/summary.html
+----
+. Open and check the Migration report (`summary.html`) at the provided URL.
```
elrodro83: I would put the `Import into Studio` step before this, but definitely before the manual part.

mule4-user-guide/v/4.1/migration-tool.adoc
```diff
@@ -0,0 +1,216 @@
...
++
+You can find all errors and warnings in the <<migration_report>>.
+. Address the errors and warnings in the report.
++
+Note that the Compatibility tool is likely work around some of these issues until your team and address them with permanent fixes.
+
+[[migration_report]]
+== Migration Report
+
+Inside each migrated project, the migration tool generates a report that contains a list of migration errors and warnings, for example:
```
elrodro83: The same info as the report is put into the mule config files as comments.

mule4-user-guide/v/4.1/migration-tool.adoc
```diff
@@ -0,0 +1,216 @@
...
+|===
+| `-destinationProjectBasePath <arg>` | Required. Directory for the migrated project.
+| `-help` | For displaying the help.
+| `-muleVersion <arg>` | Required. The Mule version to which you are migrating: `4.1.3`.
+| `-projectBasePath <arg>` | Required. Directory of the project to to migrate.
+|===
+
+== Contributions
+
+*TODO: ARE WE GOING TO LET CUSTOMERS OR 3RD PARTIES CONTRIBUTE?*
```
elrodro83: If we do, it will not be documented here anyway

mule4-user-guide/v/4.1/migration-tool.adoc
dukesphere: Good to know. I think we have to provide Beta doc. But let me double-check that.

mule4-user-guide/v/4.1/migration-tool.adoc
dukesphere: 👍

mule4-user-guide/v/4.1/migration-tool.adoc
```diff
@@ -114,21 +130,9 @@ output the results.
...
-
-*TODO: ARE WE GOING TO LET CUSTOMERS OR 3RD PARTIES CONTRIBUTE?*
-
-//TODO BELOW
-////
-* Severity (Error/Warning)
-* Description
-* Location (file:line:column)
-* Doc links: Links to useful Mule 4 documentation and the migration guide.
+Whenever the tool adds an entry to the report (either error or warning), the same information is also added as a comment in the output file. _TODO: Question: what output file?_
```
elrodro83: the mule-config-xml, or any other config files if the original had many,

mule4-user-guide/v/4.1/migration-tool.adoc
```diff
@@ -0,0 +1,313 @@
...
+| `-projectBasePath <arg>` | Required. Directory of the project to to migrate.
+|===
+
+Whenever the tool adds an entry to the report (either error or warning), the same information is also added as a comment in the output file. _TODO: Question: what output file?_
+
+== Supported Migrations
+
+The tool attempts to migrate these components, modules, and connectors:
+
+*TODO, TODO: IN PROGRESS*
```
dukesphere: @elrodro83 I went through the MMT JIRAs to take a first stab at this list (below). Can you help with this list? Note that I also listed (hidden between ////s) some that seem to be in progress or still in the TODO stage. Not sure which of those to expect any of those for the Beta.

mule4-user-guide/v/4.1/migration-tool.adoc
```diff
@@ -67,14 +36,20 @@ To run the Migration tool:
...
 +
+.Command-line Invocation
+[source,console,linenums]
 ----
 $ java -jar
- mule-migration-tool-runner-1.0.0-20180806.163412-176-jar-with-dependencies.jar
- -projectBasePath /Users/me/AnypointStudio/v6/migrator/my-v6-project
- -muleVersion 4.1.3
- -destinationProjectBasePath /Users/me/my_dir/my_migrated_project
+mule-migration-tool-runner-1.0.0-20180806.163412-176-jar-with-dependencies.jar
```
elrodro83: if this are put as separate lines in this doc, put a `\` at the end of each line but the last one, so it works ok when copy-pasted

mule4-user-guide/v/4.1/migration-tool.adoc
```diff
@@ -91,31 +66,113 @@ Migration report:
...
+
+[[compatibility_module]]
+== Compatibility Module
+
+Studio automatically applies the Compatibility module to your migrated project.
+
+You can use the module as a temporary bridge to handle some important
+functionality until you complete your migration.
+
+It is important to note that the Compatibility does not cover every migration gap. The module simply compliments of the Migration tool:
```
elrodro83: compliments -> complements

mule4-user-guide/v/4.1/migration-manual.adoc
```diff
@@ -0,0 +1,68 @@
...
+
+//TODO: REMOVE WHEN BETA IS OVER.
+*Note on the Beta Release:* This page will be augmented periodically through the
+Beta period based on feedback.
+
+
+[[migrate_mel]]
+== Migrating Remaining MEL Expressions to DataWeave
+
+When migrating MEL expressions to link:dataweave[DataWeave], you should use the
```
elrodro83: ideally, most MEL are migrated just to DW. Scripting is just for the case of procedural MEL that can't be migrated seamlessly to DW. It should be clear here that those are mot 2 equivalent options, the latter is for a very specific subset of cases

mule4-user-guide/v/4.1/migration-manual.adoc
```diff
@@ -0,0 +1,68 @@
...
+In the XML for the migrated Mule application:
+
+. Remove link:intro-mule-message#inbound-properties-are-now-attributes[compatibility:attributes-to-inbound-properties].
++
+
++
+. Replace uses of `inboundProperties` in expressions and `vars.compatibility_inboundProperties[‘propName’]`) for attributes in DataWeave
+migrations.
++
+//TODO: REPLACE WITH WHAT? NEED EXAMPLES OF BOTH.
```
elrodro83: This will depend on each connector. There should be documentation about the Attributes of each one

elrodro83: Example:

```
<http:listener config-ref="myHttp"/>
<compatibility:attributes-to-inbound-properties/>
<logger message=“#[vars.compatibility_inboundProperties[‘http.request’]]”>
<logger message=“#[mel:message.inboundProperties[‘http.request’]]”>
```
would need to be migrated to
```
<http:listener config-ref="myHttp"/>
<logger message=“#[message.attributes.requestPath]”>
<logger message=“#[message.attributes.requestPath]”>
```

mule4-user-guide/v/4.1/migration-tool.adoc
```diff
@@ -129,7 +129,7 @@ You or your team will need to handle other migration gaps that are covered in
 the <<migration_report, Migration report>>.
 
 
-//TODO: FROM RODRO: we want them to migrate to a point where the compatibility plugin can be dropped from an application
+//TODO: FROM RODRO: we want them to migrate to a point where the compatibility plugin can be dropped from an app
```
elrodro83: this is accomplished with the post-migration steps explained in the other page

mule4-user-guide/v/4.1/migration-manual.adoc
```diff
@@ -0,0 +1,88 @@
...
+*Note on the Beta Release:* This page might be augmented periodically during the
+Beta period based on feedback and recommendations.
+
+=== Using the Scripting Module to Migrate Some MEL Expressions
+
+Most MEL expressions are migrated to link:dataweave[DataWeave] by the Migration
+tool. However, for certain procedural MEL expressions that cannot be migrated
+seamlessly to DataWeave, you should use the link:scripting-module[Scripting module].
+
+//TODO: A BEFORE/AFTER EXAMPLE WOULD HELP HERE
```
elrodro83: askin g for those

mule4-user-guide/v/4.1/migration-tool.adoc
```diff
@@ -0,0 +1,372 @@
...
+*Note on the Beta release*: This list applies to the Beta and will expand over
+time.
+
+The Migration tool supports the following migrations:
+
+* Mule apps only (no domains or policies are supported)
+* MEL to DataWeave expressions (simple cases only)
+* DataWeave component to `ee:transform` (DataWeave 1.0 to DataWeave 2.0)
+* Connectors: HTTP, Database, Web Service Consumer
+* Modules: Validations
```
elrodro83: `Modules: Validations, Scripting (just DSL, no migration if the script itself)`

forgot about scritting before

ngrossi:
for me, the following statement is not clear enough... form file:///Users/nicolas.grossi/Downloads/migration-tool-topics/migration-tool.html

.........

The tool can be used for the following versions but will not be supported in the Beta or GA:

3.7.x

3.6.x

If you run the Migration tool on these unsupported apps, the migrated output might be less comprehensive than with supported apps.

..........

In some way we are saying that CAN be used but later "on these unsupported apps" :(. I guess that it is not clear for our users

ngrossi:
Another comment about the migration tool is about the command line tool name... Actually is...
.........
1 2 3 4 5 | $ java -jar mule-migration-tool-runner-1.0.0-20180806.163412-176-jar-with-dependencies.jar -projectBasePath /Users/me/AnypointStudio/v6/migrator/my-v6-project -muleVersion 4.1.3 -destinationProjectBasePath /Users/me/my_dir/my_migrated_project
-- | --
...............

But with this name "mule-migration-tool-runner-1.0.0-20180806.163412-176-jar-with-dependencies.jar" the command line execution could fail with the following message

Error: Unable to access jarfile mule-migration-tool-runner-1.0.0-SNAPSHOT-jar-with-dependencies.jar

PR #3076, opened 2018-08-08
[CS-4608]: Add documentation for identity management support with OpenID Connect by salesforce - Merge after Aug 25th 2018

shwetai:
@Mulesoft-Mysti @aj90


access-management/v/latest/conf-openid-connect-task.adoc
```diff
@@ -117,6 +117,10 @@ The URL that returns user profile information to the client app.
 * link:https://developer.okta.com/docs/api/resources/oidc.html#token-request[Okta OpenID Connect, Token Request]
 * link:https://developer.okta.com/docs/api/getting_started/getting_a_token.html[Okta Getting a Token]
 * link:https://developer.okta.com/docs/api/resources/oidc.html#get-user-information[Okta OpenID Connect, Get User Information]
+* link:https://help.salesforce.com/articleView?id=remoteaccess_oauth_web_server_flow.htm[Salesforce OAuth 2.0 Web Server Authentication Flow]
+* link:https://help.salesforce.com/articleView?id=sso_provider_openid_connect.htm[[Salesforce OpenID Connect, Authentication Request]]
```
Mulesoft-Mysti: ...connect.htm[[Salesforce... should be connect.html[, and remove the extra bracket at the end. The link works, but we see the extra brackets in the text.

Mulesoft-Mysti: One question--on line 42, we don't need to add a Salesforce api token value because they don't require registration requests be from an authorized client?

shwetai: So, refresh_token grant is supported and the documentation is regarding that. I probably should title it as [Salesforce OAuth Refresh Token Process]

access-management/v/latest/conf-openid-connect-task.adoc
```diff
@@ -102,6 +102,10 @@ The URL that provides the user’s identity encoded in a secure JSON Web Token.
 +
 The URL that returns user profile information to the client app.  
 +
+* OpenID Connect Issuer
++
+It is the the location of the OpenID Provider. Generally `/.well-known/openid-configuration` gets appended to the issuer to generate metadata URL for openID Connect specifications. For Salesforce, it is required to provide the `issuer`.
```
Mulesoft-Mysti: Delete "It is the the" and replace with The (to match other entries). 
Optionally, replace "Generally" with "For most providers," and "gets" with "is"
add "the": "to generate the metadata URL for OpenID Connect..." (capital OpenID).
Replace "it is required to" to the more active "you must provide the value for `issuer` (this also matches the voice/tense of other instructions)

PR #3081, opened 2018-08-09
RTF Docs Review

sundippatel:
Please review the documentation for Runtime Fabric. 
Leave comments, suggestions, and issues here for tracking. We'll use this to track and resolve issues brought up during the review.


nico-muley:
> ...and orchestration of your Mule applications and gateways. 
I think this should say: applications and **proxies** (?)

mcocaro:
@sundippatel Some initial comments:

*** About Anypoint Runtime Fabric ***

- Add paragraph to explain that RTF is a separate k8s cluster that does not work with BYO k8s cluster.
- Deployments of non-Mule applications or gateways on Anypoint Runtime Fabric is not supported. ---- could be more clear that non-Mule Gateways are supported in RTF.
- Connectivity to the Runtime Plane ..... we should add a paragraph that explains that the connection is SECURED through Mutual TLS over the AMQP protocol.


*** Preparing to Install Anypoint Runtime Fabric ***

- download a ZIP file containing scripts: comment that the download is from the Runtime Manager - Runtime Fabrics section of the Anypoint Platform
- This VM is called the installer VM: specify that for the purposes of the name, the installer name only works for the installation and then is just a regular controller node
- Link "Runtime Fabric as defined" gives a 404 (incorrect uri)
- Configure Log Forwarding and Alerting: How do I configure log fwd and Alerts is missing
- Environments and Inbound Traffic:
  - The certificate should require a passphrase and be set with a Common Name which is used as the domain for each deployed application on Runtime Fabric: Common Name can be a Wildcard or specific Domain. Explain Differences here.
- Access Management
  - Missing the Exchange Contributor role /// or make it clear those roles are only for the DevOps who will be installing RTF

*** System Requirements for Anypoint Runtime Fabric ***

- typo: systen level patches


*** Install Runtime Fabric on AWS ****

- Before you Begin
  - Link Review the Preparing to Install Anypoint Runtime Fabric gives a 404
  - Add a line about required Role in Access Management with a Link
- Provision Infrastructure
  - Missing AWS_DEFAULT_REGION and AWS_REGION env var that is also required
- Variables to Define
  - In general variables are not listed as REQUIRED / OPTIONAL --- not all of them are required (e.g. registration_token)
  - cluster_name -> Typo: fondrewr. Also add here limitations of the name (no special characters, length, etc)
- Running Terraform
  - Bullets are missing extra line feed
  - in the Tf-state comment, please add a line saying that each state is per cluster and that they should not re-use tf-state dir when installing multiple clusters
- Next steps
  - Missing links to Associate Env + Enable Inbound Traffic

*** Enable Inbound Traffic on Anypoint Runtime Fabric ***

- Replace Edge Configuration tab with "Inbound Traffic"
- Allocate more CPU to increase throughput and decrease latency by allocating more CPU to each replica. --- would reword to not dupe CPU 


*** Deploy a Mule Application to a Runtime Fabric ***

- Replace EDGE by Ingress // Inbound Traffic
- Missing link  [link to supported runtimes] and [link to work/core/memory topic] and [link to docs on scaling and number of cores/memory]
- Replace "Workers" by Replicas
- Missing definition of Enable inbound traffic to this application // Enable Last-Mile security // Enable application to run in Runtime Cluster Mode

mcocaro:
@sundippatel more comments ----

*** Install Runtime Fabric on Azure ***

- installation details link throws 404
- Typo: Runtime Fabric Nmae and the Example of the RTF NAme is `rtf-aws`
- Typo: defualt
- Reword the following: Your infrastructure will begin be created and configured --- missing a "to"
- Next steps missing links


*** Install Runtime Fabric Manually ***

- Typo: closet
- Vars table broken
- Format error in paragraph The output is a configur....
- Has extra data for Associate Environments to Runtime Fabric.


*** Resource Allocation on Anypoint Runtime Fabric ***

- Missing Performance metrics
- N2H: Add cluster density metrics // Sizing formula


*** Misc ***

- Missing link /deploy-about in the Index
- Missing pages in Edge Config
- Can we add the Verification of Enable Inbound Traffic on Anypoint Runtime Fabric to Deploy Apps section? or link in the Next Steps?
- Whats the difference between "Enable an Edge Endpoint for Anypoint Runtime Fabric" and "Enable Inbound Traffic on Anypoint Runtime Fabric"?
- Overall can we get rid of the README.MD files in the ZIP in favor of docs.mulesoft.com and have "installer versions" in the docs as well?


*** Missing ***

- OpsCenter docs <?> how do I set an alert? is it somewhere from PCE already?
- Terminate TLS at the Mule application + Open ports for apps (8081) and HTTPS vs HTTP in the Deploy Apps section
- IMPORTANT: Rolling application deployments with Zero Downtime
- Scaling an application
- Application Statuses

mcocaro:
*** Install Runtime Fabric in AWS ***

- Suggestion to include that users can save TF state based on cluster name *tf-data/cluster_name.state* so that they share/make changes to that state later on.

sundippatel:
Addressed issues listed in PR.

ashosborne1:
Hi team,

Docs look great, I couldn't spot any gaps in what we faced when running the first install. Im happy with this going live.

terrywang:
**About Anypoint Runtime Fabric / Determine Application Resource Allocation on Anypoint Runtime Fabric**

The terms we have been using to describe computing power (CPU, or in other words, time slices/cycles) are confusing, e.g. vCPU Cores (which is essentially CPU from k8s perspective). In `Runtime Manager Web UI - Resource allocation section`, users will see `Cores` (as CPU), and the Drag&Drop slidebar has `vCPU` (still CPU). I think we should streamline the term used for resource type - CPU (at least for Runtime Fabric).

Also, personally I'd recommend adding a bit more information on what `CPU` means for application (Runtime Fabric specific).

In k8s context, it limits and requests for CPU resources, it is equivalent to

- 1 AWS vCPU
- 1 Azure vCore
- 1 Hyperthread on a bare-metal Intel Processor with HT

It may also help to explain what `0.5` means: it guarantees half (0.5) as much CPU as one that asks for 1 CPU. The expression `0.5` is equivalent to `100m`, which can be interpreted to 100 millicpu or millicores.

> NOTE: This only applies to the container running application deployed on mule runtime (in the applicaiton pod), cluster Ops team may find other containers in the same pod with additional CPU/Memory which is not counted towards license.

nico-muley:
I think in here https://beta-rtf-design-partner-release.docs-stgx.mulesoft.com/anypoint-runtime-fabric/v/1.0/proxy-deploy-runtime-fabric 
it would be great if we could also describe how the proxy path mapping (i.e. ingress rule) will be constructed and how the API will be reachable once the proxy has been deployed.

ashosborne1:
Hi team 

Something we have noticed from discussions with HSBC today: 

We don't have anything documented about how to handle the loss/recycling of a master node. We need to document:

1) How to add a master node back into a cluster 
2) How to tell the master to redistribute the traffic across the newly added master
3) Explicitly define the failure modes when a master node goes down

sundippatel:
Known limitation w/ updating TLS configuration with the internal load balancer: users can only update with another secret in the same environment. (Cannot update with a secret in another environment.)

PR #3144, opened 2018-08-17
Mule 15363: Improve Aggregators DOC

lucianoRM:


connectors/v/latest/aggregators-documentation.adoc
```diff
@@ -9,7 +9,11 @@
 [[aggregators-reference]]
 = Aggregators Module Documentation Reference
 
-An Aggregator is a pass-through router, meaning that the same message that is processed by the aggregator is going to be processed by the components that follow it. The only modifications are the addition of some message attributes that provide information about the aggregation.
+An Aggregator is a component that receives some data, processes it to extract some value and then adds that value to a list of aggregated elements. After that, and depending on the configuration, the list of elements will be sent to a processors chain as the payload.
```
GustavoARod: I think that the usage of present tense would work better than future tense on `the list of elements will be sent to`

connectors/v/latest/aggregators-documentation.adoc
```diff
@@ -9,7 +9,11 @@
...
+
+Aggregators are pass-through routers, meaning that the same data that is received by the aggregator is going to be processed by the components that follow it. The only modifications are the propagation of variables in case any is set while executing any of the aggregation routes.
```
GustavoARod: Would worth mentioning/highlighting that this "pass-through" behavior is independent of the "...processes it to extract some value.." that means, of the expression in the `<aggregators:content>` ?

Although it might be quite rare (low chance), someone could think that the data available after the aggregator could be the same as the one that was "filtered/extracted" by the aggregator.

connectors/v/latest/aggregators-documentation.adoc
```diff
@@ -9,7 +9,11 @@
...
+
+When an aggregator releases the stored values, the list of aggregated elements will be processed either by a route within the aggregator itself or in another Mule Flow, through an aggregator listener. All of that dependes on the type and configuration of the aggregator being used.
```
GustavoARod: The same as before, I would use the present tense better

connectors/v/latest/aggregators-documentation.adoc
```diff
@@ -138,23 +148,23 @@ Aggregates elements until a time period is completed, executing the routes and l
...
 
-Every time a new event reaches the aggregator, the ID of the element will be resolved. If a group with that ID already exists in the aggregator, the value will be added to that group. Otherwise, a new group with that ID will be created and the received element will be the first element in that group aggregation.
+Every time a new element reaches the aggregator, an ID will be resolved. If a group with that ID already exists in the aggregator, the value will be added to that group. Otherwise, a new group with that ID will be created and the received element will be the first element in that group's aggregation.
 
 Some important concepts appear with the group-based aggregator:
 
-* Group timeout: When a group has to be released because all the necessary events of the group did not arrive within the expected time. If a group has timed out but is not yet evicted, it will reject attempts to add any new elements to that group.
+* Group timeout: When a group has to be released because all the necessary data for the group did not arrive within the expected time. If a group has timed out but is not yet evicted, it will reject attempts to add any new elements to that group.
```
GustavoARod: What about using elements instead of "data"

lucianoRM: I did not want to use the same word so many times.

connectors/v/latest/aggregators-documentation.adoc
```diff
@@ -138,23 +148,23 @@ Aggregates elements until a time period is completed, executing the routes and l
...
 
 * Group eviction: When a group is removed from the aggregator, regardless of whether it was completed or timed out. If a new element with that group's ID is received by the aggregator, the group will be created again.
 
-Lastly, because elements that reach group-based aggregators are related to a split sequence when the events have a `sequenceNumber`, they are sorted in increasing order prior to the aggregation release.
+Lastly, when elements that reach group-based aggregators are related to a split sequence, each with a different assigned `sequenceNumber`, they are sorted in increasing order prior to the aggregation release.
```
GustavoARod: I remember back in the time where I was developing the scenario that requires ordering I got really confused with this part. I remember that I wasn't easy to realize that this sequenceNumber was a technical detail tightly related to the Mule event. 

I think It'd be better to clarify when exactly "elements that reach group-based aggregators are related to a split sequence". Making some emphasis on the "split sequence" part

connectors/v/latest/aggregators-documentation.adoc
GustavoARod: Since this is more of a tech doc, I think It'd worth the repetitiveness in favor of promoting clarification. Anyways, it's ok to leave it like that.

connectors/v/latest/aggregators-documentation.adoc
```diff
@@ -192,7 +205,7 @@ Lastly, because elements that reach group-based aggregators are related to a spl
 
 `<aggregators:aggregator-listener aggregatorName="exampleAggregator" includeTimedOutGroups="false">`
 
-Once the aggregator that is referenced by the listener completes an aggregation, the listener will be triggered with a list of all the elements. Though the aggregation listener can be used for any kind of aggregator, it is important for time-driven aggregations. Such aggregations are triggered asynchronously and outside an event context, so they do not execute an aggregator route and can only reach components in flows with an aggregator listener as the source.
+Once the aggregator that is referenced by the listener completes an aggregation, the listener will be triggered with a list of all the elements. Though the aggregation listener can be used for any kind of aggregator, it is important for time-driven (Async) aggregations. Such aggregations are triggered asynchronously, so they do not execute an aggregator route and can only reach components in flows with an aggregator listener as the source.
```
GustavoARod: Present tense would be better

connectors/v/latest/aggregators-documentation.adoc
```diff
@@ -213,34 +226,49 @@ Each time a message goes through an aggregation, some attributes with informatio
...
 |======================
 
-== Time-Based Versus Event-Driven Aggregations
-
-There are two kinds of triggers for aggregations: synchronous (or event-driven) and asynchronous (or time-driven). As seen in the configurations, an aggregation can be considered complete based on a new value being added to the list (as when a max size is specified) or because some timeout or time period was completed. _This is important because the type of aggregation determines which chain of components to execute with that list of elements._
+== Async vs Sync Aggregations
 
-Each time counter associated with an aggregator starts counting the moment the first message of a group arrives. Once the aggregation is complete, the counter resets and waits for the next element to arrive.
+There are two kinds of triggers for aggregations: Synchronous and Asynchronous. +
+As seen in the configurations, an aggregation can be considered complete based on a new value being added to the list (as when a max size is specified) or because some timeout or time period was completed. _This is important because the type of aggregation determines which chain of components to execute with that list of elements._
```
GustavoARod: 1. I would go for changing, value by element
2. _"..to execute with **that** list of elements"_  --> ...to execute with **the** list of elements

connectors/v/latest/aggregators-documentation.adoc
```diff
@@ -213,34 +226,49 @@ Each time a message goes through an aggregation, some attributes with informatio
...
 
+[IMPORTANT]
+Each time counter associated with an aggregator starts counting from the moment the first message of a group arrives. Once the aggregation is complete, the counter resets and waits for the next element to arrive. +
 For single-group aggregators (time-based and size-based aggregators), there is only one time counter, but for the group-based aggregator, there is one counter per group.
 
-If an aggregation is released by a time period or timeout completion, it will never execute the routes in its definition. That leads to an unwanted scenario where only a piece of a flow is executed, starting from the processor's chain configured inside the aggregator and continuing with the components that follow it but without executing any from the flow's source to the aggregator itself.
+For aggregations that complete because a new element arrives (Sync), at least one of the following things will happen:
```
GustavoARod: ..aggregations **completed**

GustavoARod:
I really like the changes. Nevertheless, I think it worth enhancing the information about the object-store relation with aggregators and its implications, like:

Some of the ideas I remember from my experience with the review are. 
1. Persistent OS implies high reliability
2. Persistent OS implies serialization when dealing with Java data types.
3. The default OS config is a persistent one
4. For configuring a new OS through the Aggregators properties UI. OS module dependency needs to be already added to the project

lucianoRM:
I think those are OS related explanations and are valid for any connector that persists data in an objectStore. 
However, a clarification should be added related to the fact that any aggregated content should be serializable. And maybe add a link the the OS documentation. The problem would be if there isn't any

PR #3203, opened 2018-08-29
DOCS 2702 for Get Well

rddheath:


rddheath:
As we've talked about, rather than giving examples in the topics for these components, I thought it better to point to the example projects in Exchange. Otherwise, when the example projects are updated or revised, we'll have to revise the readme.md files for the projects and the examples based on them that are in the documentation. Moreover, because developers have come up with example projects, it makes sense for us to use them, rather than work with potentially the same developers to come up with different examples just for the topics.

mule4-user-guide/v/4.1/cache-scope.adoc
```diff
@@ -55,6 +55,10 @@ REMOVED:
 
 Instead of processing all message payloads that it receives, the Cache scope can exclude specific payloads from the Cache scope flow based on an DataWeave expression.
 
+== Example
+
+In Anypoint Studio, you can download and open the example project _Cache Scope with Salesforce Contacts_ from Anypoint Exchange to learn more about how to use the Cache scope. This example caches fetched data from Salesforce. When a query is issued to request the data again, results are returned from cache and a new call to Salesforce is not made.
```
Mulesoft-Mysti: Why don't we link to the example project? It's 3-5 steps to get there from docs.mulesoft.com. I'd expect a link at least in the See Also.

mule4-user-guide/v/4.1/scatter-gather-concept.adoc
```diff
@@ -9,6 +10,27 @@ image::scatter-gather-concept-56ccc.png[]
 
 Scatter-Gather executes routes concurrently instead of sequentially. Parallel execution of routes greatly increases the efficiency of your application and provides more information than sequential processing.
 
+In Anypoint Studio, you can download and open the example project _Scatter-Gather Flow Control_ from Anypoint Exchange to learn more about how to use the Scatter-Gather component. This example shows the usage of the scatter-gather control flow to aggregate data in parallel and return the result in JSON.
+
+The example uses prepared data as input for two resources that should be aggregated. The data represents information about two contacts and has the following structure:
```
Mulesoft-Mysti: Same comment as above--why not link to the example?

mule4-user-guide/v/4.1/for-each-scope-concept.adoc
```diff
@@ -30,16 +30,18 @@ You can also split an array in batches. Potentially, batches produce quicker pro
...
-Importing Email Attachments using the POP3 Connector
-Querying a Database and Attaching Results to an Email
+There are several example projects in Anypoint Exchange that you can open in Anypoint Studio to learn more about how to use the For Each scope:
+
+* _Authenticating Salesforce using OAuth2_
+* _Import contacts into Microsoft Dynamics CRM_
+* _Importing a CSV file into mongoDB_
+* _Importing an Email Attachment using the IMAP Connector_
+* _Importing Email Attachments using the POP3 Connector_
+* _Querying a Database and Attaching Results to an Email_
```
Mulesoft-Mysti: Why don't we link to these examples?

mule4-user-guide/v/4.1/cache-scope.adoc
rddheath: Because, to open the project in Studio, the user has to go to Exchange from Studio. I didn't think it served users well to tell them to open a project from Studio and give them a link, implying they could open the project at the link.

rddheath: Also, one has to be logged into Anypoint Platform for the link to work. If one isn't logged in, the link will open the login page and then open Exchange, but will not display the project page in Exchange.

mule4-user-guide/v/4.1/scatter-gather-concept.adoc
rddheath: For the same reasons that I list for your previous comment.

mule4-user-guide/v/4.1/cache-scope.adoc
Mulesoft-Mysti: Why do we assume they are or want to be in studio to look at the example? If that is a valid assumption, then can we rephrase to remind them where/how to search (without a lot of verbiage)? It just feels like a dead end as is.

rddheath: I'm not assuming users will want to be in Studio or that they currently are in Studio. 
1. Unless users are already logged into Anypoint Platform when they click the link, the link will simply open the login page for Anypoint Platform. After users log in, they will be taken to Exchange, but not to the example project. They will still have to search for the example project by name. 
2. Suppose that users are logged into Anypoint Platform when they click the link. Then, they will be able to read the description of the example, but they will still need to open the example project from Studio. As they can read the description also when they open Exchange from Studio, it doesn't seem useful to put in a link that only a subset of users are likely to follow successfully and that, even if they do follow it, won't allow them actually to open the example.

If you're rationale for putting in a link is that a link is better than nothing, I disagree. It's worse than nothing because these two reasons ensure a poor user experience.

rddheath: I have, however, added text to describe how to open an example project in Studio.

Mulesoft-Mysti: Perfect!

rddheath: Would you approve the changes, then? I can't merge until then.

PR #3217, opened 2018-08-31
Exchange Categories doc fixes

george-hoffer:
From slack discussions

### DIFF
--- a/anypoint-exchange/v/latest/to-manage-categories.adoc
+++ b/anypoint-exchange/v/latest/to-manage-categories.adoc
@@ -9,6 +9,9 @@ and `Integration API` is applied to assets of type REST API, SOAP API, and HTTP 
 
 image:ex-categories-api-type.png[Settings Category for API Type]
 
+*Note:* You must have one or more categories defined in the Settings page for the Add Category prompt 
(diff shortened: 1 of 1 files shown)


anypoint-exchange/v/latest/to-manage-categories.adoc
```diff
@@ -46,7 +49,8 @@ The following search terms return all assets marked as both System APIs and Sale
 == To Change or Delete a Category
 
 After you apply a category to an asset, to change or delete a category from the
```
kuntald: This whole section seems a bit confusing. I'm adding @antonkravchenko as well, to make the final decision. THe key message is that categories can be edited/deleted from two levels: 1) At the individual asset level, without impacting other assets, and 2) at the "global" settings level, where users can delete the whole category itself, which will in turn delete the particular category from ALL assets. 

It wasn't very clear from the two paragraphs. Is there any way you can simplify the statements. Maybe just say something along the line of "Categories can be deleted at the individual asset level which will only remove the category from that asset. Categories can also be deleted at the global level in Settings. But doing so will remove the category from all assets." ... But probably with better words & english than what I wrote ;).

PR #3223, opened 2018-09-03
Clarified default values

PabloCamino:
Made more enfasis on the need to have this for metadata resolution to work

### DIFF
--- a/mule4-user-guide/v/4.1/secure-configuration-properties.adoc
+++ b/mule4-user-guide/v/4.1/secure-configuration-properties.adoc
@@ -141,7 +141,7 @@ This way, the value of the property `env` determines which file to use to load t
 <secure-properties:config key="${runtime.property}" file="${env}-properties.yaml" name="test"/>
 ----
 
-This way, the default value for the `env` property will be `"dev"`, which can still be overriden with a system or environment property. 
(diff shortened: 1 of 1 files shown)


mule4-user-guide/v/4.1/secure-configuration-properties.adoc
```diff
@@ -141,7 +141,7 @@ This way, the value of the property `env` determines which file to use to load t
 <secure-properties:config key="${runtime.property}" file="${env}-properties.yaml" name="test"/>
 ----
 
-This way, the default value for the `env` property will be `"dev"`, which can still be overriden with a system or environment property. 
+This way, the default value for the `env` property will be `"dev"`, which can still be overriden with a system or environment property. Please note that this is *required* for metadata resolution in Anypoint Studio. If you do not define devault values for the proerties that are passed through the command line you will receive an error while creating application model.
```
Mulesoft-Mysti: proerties should be properties.

Mulesoft-Mysti: "while creating" should probably be "when you create the"

PabloCamino:
Changed proerties. I left the other writing as it's the exact same frase that appears on Anypoint Studio.  The application model is not something the user creates but built by the program in the background to expose the metadata.

PR #3260, opened 2018-09-09
MetaDataRetriever return type miss match

RanveerMeel:
Return type of MetaDataRetriever is metaData and it is returning DefaultMetaData so type miss match.
I have just added 1 line of code.


george-hoffer:

RanveerMeel:
I have done as per your mail. Please check.

PR #3290, opened 2018-09-12
DOCS-2879 Changes

rddheath:
The entire topic presented the table as a list of recommendations, not hard limits. Therefore, I couldn't simply change the heading of the second column in the table. I had to change all uses of the term "recommendations".

### DIFF
--- a/runtime-manager/v/latest/runtime-dep-serv-limits.adoc
+++ b/runtime-manager/v/latest/runtime-dep-serv-limits.adoc
@@ -1,20 +1,17 @@
-= About Deployment and Server Scaling Recommendations
+= About Hard Limits on Scaling Deployments and Servers
 
-This topic describes recommendations on scaling the number of deployments and servers you have in a hybrid deployment scenario.
+To ensure platform stability and performance, Anypoint Platform has set the following hard limits on the number of deployments and servers that you can have in a hybrid deployment scenario. These hard limits apply to individual environments in a business group.
 
-To ensure platform stability and performance, Anypoint Platform has the following recommended limits on the number of deployments and servers. These recommendations are specific to an environment in a business group. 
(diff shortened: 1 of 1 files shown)


runtime-manager/v/latest/runtime-dep-serv-limits.adoc
```diff
@@ -1,20 +1,17 @@
-= About Deployment and Server Scaling Recommendations
+= About Hard Limits on Scaling Deployments and Servers
```
mcocaro: the currently imposed quotas are not Hard limits but rather soft limits, we have an internal KB for what the hard limits are.

runtime-manager/v/latest/runtime-dep-serv-limits.adoc
```diff
@@ -1,20 +1,17 @@
...
 
-This topic describes recommendations on scaling the number of deployments and servers you have in a hybrid deployment scenario.
+To ensure platform stability and performance, Anypoint Platform has set the following hard limits on the number of deployments and servers that you can have in a hybrid deployment scenario. These hard limits apply to individual environments in a business group.
```
mcocaro: this needs to be worded out to explain the limits for Hybrid Standalone servers and Runtime Fabric.

runtime-manager/v/latest/runtime-dep-serv-limits.adoc
rddheath: Was the SE ticket for the requested changes incorrect? https://www.mulesoft.org/jira/browse/SE-9016

runtime-manager/v/latest/runtime-dep-serv-limits.adoc
rddheath: This wasn't a requirement for the SE JIRA ticket, the URL to which I posted in my reply to your previous comment. Would you open a separate DOCS JIRA ticket for this and assign it to Steven?He's the primary owner for the Runtime Manager documentation. I'm filling in to close the SE ticket.

PR #3296, opened 2018-09-13
DOCS-2821 raw data, DOCS-2819 tools, DOCS-2481 logs download

dukesphere:


dukesphere:
@pgaliana This covers Raw Data and Tools. If someone else should review it, pls let me know. Can you do this before Monday? Thanks!

PR #3297, opened 2018-09-13
Crowd APIM perms note after user classifies APIs

sharkberto:
Note for user how to grant permissions to environments after APIs are classified


george-hoffer:
Also added to new repo on https://github.com/mulesoft/docs-release-notes/pull/5

PR #3300, opened 2018-09-13
Document new Java module error functions and add details on migration…

afelisatti:
… guide.

### DIFF
--- a/connectors/v/latest/java-throwable.adoc
+++ b/connectors/v/latest/java-throwable.adoc
@@ -7,15 +7,13 @@ toc::[]
 
 The Java module exposes useful functions to deal with Throwable instances.
 
-IMPORTANT: These functions are available since version 1.2.0
+IMPORTANT: These functions have been available since version 1.2.0.
 
 == Checking for a Cause
 
 You can check whether a Throwable was caused by a certain type of Exception using the `isCausedBy` function.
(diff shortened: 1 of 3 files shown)


connectors/v/latest/java-throwable.adoc
```diff
@@ -0,0 +1,65 @@
...
+    </on-error-continue>
+    <on-error-continue type="#[Java::isCausedBy(error.cause, 'java.lang.NullPointerException')]">
+        <set-payload value="There is no price."/>
+    </on-error-continue>
+</error-handler>
+----
+
+== Obtaining a Root Cause
+
+You can also obtain the root cause of a Throwable, going through all the possible wrapping, with the `getRootCause` function.
```
dukesphere: Re "going through all the possible wrapping"? Are you talking about going through all possible wrapped expressions?

afelisatti: No, Java exception wrapping, meaning when one exception is the cause of another and another and so on. This function gets the root, original one.

dukesphere: Yikes, I meant "wrapped Java exceptions," not "expressions." Anyway, "going through all the possible wrapping" sounds strange. Can you reword it?

PR #3324, opened 2018-09-17
DOCS-2903 remove references to visualizer viewer permission

yaise:


sleslie23:
See also: https://github.com/mulesoft/docs-visualizer/pull/1

